PR #2, opened 2022-11-18
initial docs

DT-DawidWroblewski:
Initial simSwap API // ATP documentation


documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,140 @@
...
+|HTTP Header|Usage Category|Description|
+|:---|:---|:---|
+|User-ID-Type|REQUIRED [if the bearer type access token provided in the request is generic]|MC User identification type used in the User-ID header. One of these values MUST be used: <ul><li>**MSISDN**: Indicates the User-ID contains a plain MSISDN in international format according to ITU-T recommendation E.164. The plus sign (+) MUST NOT be included as a prefix.</li><li> **ENCR_MSISDN**: Indicates the User-ID contains an encrypted MSISDN. Support for the MSISDN user ID type is REQUIRED, whereas for ENCR_MSISDN is OPTIONAL.</li><ul>|
+|User-ID|REQUIRED [if the bearer type access token provided in the request is generic]|MC User identification value as per the type indicated in the HTTP header above.|
+
+Both the names and the values of these HTTP headers MUST be treated as case insensitive.
+
+### ATP API details
+
+YAML proposal:
```
jlurien, line 85: We are open to discuss both proposals but TEF view is more aligned with Orange approach of having separate endpoints, and considering three_legged tokens with dedicated scopes that may require explicit consent. However for us MSISDNs must not be required as input for the API.

jlurien (commented):
TEF view on this API is quite compatible with the proposal presented by Orange. As commented by my colleague Rubén on the commonalities issue [#105](https://github.com/camaraproject/WorkingGroups/issues/105), we are thinking about an API with dedicated endpoints to get SIM Swap-related information, not having a generic /userinfo as in MobileConnect, with this information mixed with other user information. This is in line with the approach followed for other CAMARA APIs.

Those endpoints would be protected by scopes as well, and an access_token providing access to those scopes, is required prior to invoke the endpoints, so that part is quite similar. However, after checking with our Legal people, it seems that explicit user consent is probably required, so we foresee the necessity to consider three_legged tokens. When using three_legged tokens, as they can be related to the MSISDN, there is no need to include this information explicitly as part of the request, in the body or using headers. For TEF, MSISDNs are considered sensible information and they must not be required as input for the API. 

Regarding the API functionality, currently our requirements only demand an endpoint to get the timestamp of the last swap, but it would be OK to have a separate endpoint to get a bool based on a threshold, as implementation effort is similar.

jlurien:
Thanks @DT-DawidWroblewski,

I'm familiar with MobileConnect and I do not doubt that it serves for the use case. But the approach of having dedicated endpoints instead of a single one works as well, and the objective of CAMARA is to agree on the better way to offer the functionality to developers, even if it means to move from what we have right now. This debate is not specific to this API and we probably need to resolve it via commonalities or at another level.  

Regarding the need to provide MSISDNs, as you say, it is required *as long as the bearer type access token provided in the request is generic*. That takes us to the matter of using client_credentials vs three_legged. Here, it depends ultimately on what is the purpose to use the API, not on the API itself. I leave interpretation of GDPR to lawyers, but GDPR is the same across EU, so I expect we should reach a common understanding at some point. Nevertheless, we don't have to restrict it only to client_credentials in any case. For sure, if client_credentials are considered, an explicit way to identify MSISDNs is required as input, and there we can discuss options, such as allowing clear MSISDNs or having another related id. This is also a debate ongoing in commonalities.

code/API_definitions/checkSimSwap-v0.3.0.camara.swagger.yaml
```diff
@@ -0,0 +1,253 @@
...
+  version: 0.3.0
+externalDocs:
+  description: Product documentation at Camara
+  url: https://github.com/camaraproject/
+security:
+  - oAuth2ClientCredentials: []
+  - BasicAuth: []
+  - apiKey: []
+  - three_legged:
+    - simSwap-read
```
jlurien, line 28:
```suggestion
    - sim-swap-read

```

code/API_definitions/checkSimSwap-v0.3.0.camara.swagger.yaml
```diff
@@ -0,0 +1,253 @@
...
+servers:
+  - url: "{apiRoot}/{basePath}"
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+      basePath:
+        default: simSwap/v0
```
jlurien, line 36:
```suggestion
        default: sim-swap/v0

```

code/API_definitions/checkSimSwap-v0.3.0.camara.swagger.yaml
```diff
@@ -0,0 +1,253 @@
...
+        description: Base path for the simSwap API
+paths:
+  /retrieveSimSwapDate/:
```
jlurien, line 39: In this case a GET /last-sim-swap?msisdn=XXXXXX may be more RESTful, as we are only accessing info

code/API_definitions/checkSimSwap-v0.3.0.camara.swagger.yaml
```diff
@@ -0,0 +1,253 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+              example:
+                code: INTERNAL
+                message: "Session could not be created"
+        "503":
+          $ref: "#/components/responses/Generic503" 
+  /checkSimSwap:
```
jlurien, line 84: /check-last-sim-swap

code/API_definitions/checkSimSwap-v0.3.0.camara.swagger.yaml
```diff
@@ -0,0 +1,253 @@
...
+        authorizationCode:
+          authorizationUrl: https://auth.example.com/authorize
+          tokenUrl: https://auth.example.com/token
+          scopes:
+            simSwap-check: check simSwap
+  schemas:
+    SimSwapInfo:
+      type: object
+      properties:
+        sim_change:
```
jlurien, line 163: last_sim_swap

code/API_definitions/checkSimSwap-v0.3.0.camara.swagger.yaml
```diff
@@ -0,0 +1,253 @@
...
+          type: string
+          description: Code given to this error
+        message:
+          type: string
+          description: Detailed error description
+    CreateCheckSimSwap:
+      type: object
+      required:
+        - msisdn
+        - age
```
jlurien, line 195: max_age

jlurien, line 195: If a default is defined, there is no need to make it required

jlurien (commented):
We are quite aligned with the proposal by Orange. I made some editorial suggestions for wording or  to adapt it to the recently approved guidelines  in Commonalities (/path-resource, GET vs POST...).

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,140 @@
...
+![Figure 1 - Mobile Connect Client Credentials Mode Flow](assets/images/figure1cc.png)
+
+The high level flow is as follows:
+
+- The SP needs an Access Token not tied to the user for a specific use case. This assumes that the SP is already registered with Mobile Connect.
+- The SP sends an Access Token request to the ID GW (Authorization Server) with a grant type “client_credentials” and optional scope(s) (representing the resource(s) requested).
+- The ID GW authenticates the SP using its client credentials. If the request is valid, the ID GW sends an Access Token for the scope(s).
+- Where requested, the SP can then call the relevant resource endpoint (*/userInfo*) by submitting the received Access Token to retrieve the requested attributes/claims (not shown in Figure 1).
+
+### User Consent Management
```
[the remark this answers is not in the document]
DT-DawidWroblewski: Hi! Sure, this makes sense, although this should happen inside "authentication" call, not inside during a resource endpoint call, that is expected to be the last one in a call flow. Mobile Connect Client Credentials Profile flow matches customer demand, where no explicit consent is being captured by MNO. However, the Mobile Connect framework allows to have it:

- over a redirect (OIDC Device Initiated) to a target that can capture the consent (HTML page, SMS, other authenticators)
- over a push coming from a server (CIBA), e.g. by using SMS

documentation/API_documentation/API_definition.md
DT-DawidWroblewski: Currently MSISDN is required as usually it is the only identifier passed by the user, that can be passed to MNO for resolution.
**Generic userinfo endpoint** works for Mobile Connect framework where developer/SP plays only with scopes to distinguish API service.

OIDC Mobile Connect Device Initiated follows **three_legged tokens** approach, so we're fine here. This works perfectly for Number Verify, although what customers desires is to work with **delegated consent** captured by SP, instead of MNO involved inside User Journey (e.g. by HTML pop-up to request access grant).

bigludo7 (commented):
Thanks @DT-DawidWroblewski 
Reviewed with Patrice
We're making together huge progress 

Few comments:

- Why all theses attributes in header response (pragma, content-type etc...) - we should have the minimum information to avoid confusion.
- We need probably to have 3 scopes explicitly defined in the swagger: mc_atp, simswap_date, simswap_verify
- We need an 'age' attribute for simswap/verify as mandatory request parameter (Delay in hours to be checked for simswap).

Thanks
Ludovic/Patrice

DT-DawidWroblewski:
Hi!


  *   These headers are postman artifacts – we can delete them.
  *   Scopes: fine for me
  *   Age (I assume in hours) – this his tricky and I’d skip it for now. Maybe we can have this differentiated on a scope level and have sth like this: “sim_swap_verify_24” to check if simSwap took place in last 24 hours and “sim_swap_verify_48” to check if simSwap took place in last 48 hours. I assume that this parameter differs between countries (there is no common legal recommendations for this) and having a solution where developer gets errors while looping over “hour” attribute to check which one works, doesn’t make sense.


Regards,
David


Deutsche Telekom
Technology Delivery International

Dawid Wróblewski
Marynarska 12, 02-674 Warsaw
+48 602 204 473 (phone)
Email: ***@***.******@***.***>
www.telekom.com<http://www.telekom.com/>

***@***.***


Life is for sharing.

You can find the obligatory information on www.telekom.com/compulsory-statement<http://www.telekom.com/compulsory-statement>

bigludo7:
Thanks @DT-DawidWroblewski 
Not a big fan of using scope for age because it depends on country but also probably on UC. Adding scope will require API new version and so deployment.
Having an age attribute is simplest for us - why is it tricky from your perspective?
Thanks

DT-DawidWroblewski:
Hi!

Tricky because e.g. in Germany it can be 24h, in France 6h, in Spain 48h.

I mean – documentation can carry this information, but then we’re creating market/country specific deployment.

So message to developer would be sth like this:

Spain: allowed age 48
France: allowed age 6
Germany: allowed age 24
…
{Country X}: allowed age 480

If you’re developing an app that works worldwide you need to maintain configuration per country… alternatively looping over “age” attribute to finally get answer. If you’re not by throttling mechanisms, you’re lucky 😉

Regards,
David


Deutsche Telekom
Technology Delivery International

Dawid Wróblewski
Marynarska 12, 02-674 Warsaw
+48 602 204 473 (phone)
Email: ***@***.******@***.***>
www.telekom.com<http://www.telekom.com/>

***@***.***


Life is for sharing.

You can find the obligatory information on www.telekom.com/compulsory-statement<http://www.telekom.com/compulsory-statement>

monamok:
Hello,
as @jlurien is OOO, I'm reviewing this API on his behalf. Sorry if I'm asking about some points that you already talked about them.

I have a question about the scopes that you're mentioning @bigludo7:
> We need probably to have 3 scopes explicitly defined in the swagger: mc_atp, simswap_date, simswap_verify

As far as I'm reading here, there's gonna be two endpoints, one to return last_swap (simswap_date scope) and another one to verify if it's been swapped (simswap_verify scope). What is mc_atp? 

@DT-DawidWroblewski I saw that you're editing the original proposal but @jlurien were commenting on the second proposal so I'm not sure which one of two files we're talking about. Are you going to remove /token and /openid/userinfo endpoints from your version eventually?

bigludo7:
Hello @monamok 

> I have a question about the scopes that you're mentioning @bigludo7:
> 
> > We need probably to have 3 scopes explicitly defined in the swagger: mc_atp, simswap_date, simswap_verify
> 
> As far as I'm reading here, there's gonna be two endpoints, one to return last_swap (simswap_date scope) and another one to verify if it's been swapped (simswap_verify scope). What is mc_atp?
> 

In @DT-DawidWroblewski proposal we have 3 endpoints: one for 'legacy' MC Account takeover Protection (the userinfo), one for simSwap providing a date and one for simswap providing a boolean.

From my perspective one given implementation has not to provide these 3 flavors (depending on its capacity) but at least one.

monamok:
Thank you @DT-DawidWroblewski and @bigludo7 for the answers. About this part:

> currently, we agree to have 3 endpoints for protected resources in simSwap family:

> /userinfo
> /simSwap/verify
> /simSwap/last

> mc_atp scope goes for /userinfo endpoint

> we can take discussion about /token endpoint to commonalities, but I'm keeping swagger that holds the whole flow here as it >looks that it is different than what we're discussing in commonalities.

I think we already stablished in the last meeting that getting the token should be handled outside of the API. Please let's keep it simple and stick to the second and third endpoints that were included in Orange proposal. 
And we also talked about having specific resourses before/during/after the meeting. in /userinfo response there are some fields that are already part of the two other endpoints, so they're redundant and there are parameters like "accountState" that by itself might be interesting data but they are out of scope of Sim Swap API. Could you please reconsider it?

DT-DawidWroblewski:
Due to the fact that simSwap information is related to the User's identity and Mobile Connect APIs utilize standardised **/userinfo** endpoint, I recommend keeping 3 endpoints (btw /userinfo exists at OpenID Forum [OpenID Forum Link](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo), TM forum [TMforumLink](https://www.tmforum.org/resources/specification/tmf691-federated-identity-api-user-guide-v4-0-0/#) as a standardized way to provide information about user identity. Therefore, I recommend sticking to the standards (leave /userinfo), but also introducing alternative ways to consume simSwap information)

MarkusKuemmerle:
@DT-DawidWroblewski : Please contribute API definitions in the code/API_definitions/ subdirectory and not in code/API_code/. Please change this pull request in this way.

PR #5, opened 2022-12-12
initial documentation ONLY

DT-DawidWroblewski:


DT-DawidWroblewski:
I removed yaml files, so we can work on them inside a new PR

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,140 @@
...
+Content-Length: 0
+
+#### Resource Response
+
+- Boolean
+>{
+    "simChange":"false"
+    "correlation_id":"6d9d6d48-b01b-4460-a9f0-d696bcdfeb95"
+}
+- Date
```
bigludo7, line 131: Why adding this flavor ?  for me the boolean and the timestamp were just good enough

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,140 @@
...
+Transfer-Encoding: chunked
+Connection: keep-alive
+Cache-Control: no-store
+Pragma: no-cache
+Strict-Transport-Security: max-age=15724800; includeSubDomains
+{"access_token":"{AccessToken}","scope":"mc_atp","token_type":"Bearer","expires_in":"500"}
+
+#### Resource Request
+
+>POST /openid/userinfo HTTP/1.1
```
bigludo7, line 114: We discussed also 2 other that should be there resources 

POST /userinfo for MC_ATP
POST /simSwap/verify for boolean  as in deviceStatus API
POST /simSwap/last to serve timestamp

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,140 @@
...
+|HTTP Header|Usage Category|Description|
+|:---|:---|:---|
+|User-ID-Type|REQUIRED [if the bearer type access token provided in the request is generic]|MC User identification type used in the User-ID header. One of these values MUST be used: <ul><li>**MSISDN**: Indicates the User-ID contains a plain MSISDN in international format according to ITU-T recommendation E.164. The plus sign (+) MUST NOT be included as a prefix.</li><li> **ENCR_MSISDN**: Indicates the User-ID contains an encrypted MSISDN. Support for the MSISDN user ID type is REQUIRED, whereas for ENCR_MSISDN is OPTIONAL.</li><ul>|
+|User-ID|REQUIRED [if the bearer type access token provided in the request is generic]|MC User identification value as per the type indicated in the HTTP header above.|
+
+Both the names and the values of these HTTP headers MUST be treated as case insensitive.
+
+### ATP API details
+
+YAML proposal:
```
bigludo7, line 85: We should remove this as we can hope to have one yam with 3 endpoints ?

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,140 @@
...
+Many online banking accounts today are protected using secure online banking credentials (e.g., password and secret customer number), and additionally using the phone number associated with the User’s bank account to increase security. This two-factor security helps mitigate fraud if one of the factors is compromised, for example if the User’s online banking credentials have been stolen.
+Mobile Connect ATP allows SPs to perform additional checks on the User’s mobile phone status to determine whether or not this factor may have been compromised.  In particular, the service returns an indication of whether there has been a recent SIM swap and may optionally return other information relating to the User’s mobile account:
+
+- Recent SIM change (MSISDN - IMSI pairing) - REQUIRED
+- Active unconditional call diverts - OPTIONAL
+- Device reported lost or stolen - OPTIONAL
+- Recent device change (Timestamp of last MSISDN - IMEI pairing change) - OPTIONAL
+- Mobile phone account status (active or inactive) – OPTIONAL
+
+## Use Case examples
```
bigludo7, line 28: We should probably explicitly make the difference between the Camara SimSwap API 'core'  that only focus on simswap from the MC ATP API which has a larger scope. One way to implement the Camara SimsWap API is to follow MC ATP but as we introduced 2 other endpoint strictly pointing on simswap (with a boolean or a timestamp) .

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,140 @@
...
+
+Mobile Connect ATP offers a range of practical use cases. Some of the use cases identified for implementation are listed in the table below:
+
+1. Prevention of various types of fraud in banking such as SIM swap fraud
+2. Making SMS one-time passwords more secure
```
bigludo7, line 33: Not in this API

bigludo7 (commented):
Hi
This documentation is a bit confusing because it is mainly about MC ATP and not the discussed Camara simSwap API.  Look for example the resource request line 112 that only cover userinfo and not the 2 other endpoints we discussed.
I have added directly comment in the file itself.
Thanks

Ludovic

DT-DawidWroblewski:
Hi!

Thx – I’ll fix it and add changes/remarks.

Regards,
David


Deutsche Telekom
Technology Delivery International

Dawid Wróblewski
Marynarska 12, 02-674 Warsaw
+48 602 204 473 (phone)
Email: ***@***.******@***.***>
www.telekom.com<http://www.telekom.com/>

***@***.***


Life is for sharing.

You can find the obligatory information on www.telekom.com/compulsory-statement<http://www.telekom.com/compulsory-statement>

documentation/API_documentation/API_definition.md
DT-DawidWroblewski: This is TimeStamp flavor. In some countries we received negative feedback from Data Privacy teams to provide as detailed info as TimeStamp (with hours, minutes, seconds).

jlurien (commented):
I think we have to agree on a common YAML file first, then we can adjust documentation accordingly. Here, I don't see we are moving forward from the blocking points as this documentation is still focused on Mobile Connect and is hard to review the API proposal itself

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,201 @@
+# API Reference
+
+CAMARA Sim Swap is based on Mobile Connect Account Takeover Protection (ATP) standardized product:
```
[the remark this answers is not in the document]
DT-DawidWroblewski: Hi!

I don't see it this way.

CAMARA & Mobile Connect APIs have a lot in common.

The fact that CAMARA utilizes Mobile Connect should not be considered controversial or bad - it is rather utilizing assets/solutions that are available on a market.

CAMARA follows all concepts that are described inside MC specs, just to mention some:
1. MC & CAMARA follows OpenID Connect & OAUTH2.0 frameworks
2. MC & CAMARA use scopes to protect APIs
3. MC & CAMARA shows flexibility for naming the endpoints that handle resource API calls

However, CAMARA lacks details that are already described under the Mobile Connect framework, like:
1. CAMARA has **no focus** on OAUTH2.0 Client Credentials flow, as most APIs supports only OAUTH2.0 **code flow** (3-legged-tokens). Client Credentials enable simple and fast API consumption.
2. CAMARA end-user authentication concept is still under discussion, whereas MC has it well documented and described.
3. CAMARA out scopes Identity APIs and focus more on APIs outside the user context, whereas User Identity and related attributes are in the scope of MC APIs - this is why **/userinfo** endpoint is recommended way to consume Identity-related APIs.

During discussions I'm having here I see it as less controversial, but rather more in common. The most difficult thing I find is that MC specs cover the whole API flow (from /authorize to resource endpoint call), whereas CAMARA separates API design into commonalities and resource calls.

This remark also touches [Issue#105](https://github.com/camaraproject/WorkingGroups/issues/105), so I'm in favour of closing that one, as I find more and more things that are common between CAMARA and MC.

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,201 @@
...
+
+##### Resource Request - simSwap verify
+
+>POST /simSwap/verify HTTP/1.1
+Content-Type: application/x-www-form-urlencoded
+Authorization: Bearer {AccessToken}
+Host: server.example.com
+Accept-Encoding: gzip, deflate, br
+Connection: keep-alive
+Content-Length: 0
```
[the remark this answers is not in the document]
DT-DawidWroblewski: This can be enabled by passing a proper value inside scope parameter, e.g. scope=simswap_verify_24

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,201 @@
...
+**Figure 1** illustrates the Client Credential mode flow. This specification details the parameters involved in the Access Token Request and Response.
+
+![Figure 1 - Mobile Connect Client Credentials Mode Flow](assets/images/figure1cc.png)
+
+The high level flow is as follows:
+
+- The SP needs an Access Token not tied to the user for a specific use case. This assumes that the SP is already registered with Mobile Connect.
+- The SP sends an Access Token request to the ID GW (Authorization Server) with a grant type “client_credentials” and optional scope(s) (representing the resource(s) requested).
+- The ID GW authenticates the SP using its client credentials. If the request is valid, the ID GW sends an Access Token for the scope(s).
+- Where requested, the SP can then call the relevant resource endpoint (*/userInfo*) by submitting the received Access Token to retrieve the requested attributes/claims (not shown in Figure 1).
```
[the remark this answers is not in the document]
DT-DawidWroblewski: /userinfo endpoint is used for MC ATP flavor of simSwap API family

documentation/API_documentation/API_definition.md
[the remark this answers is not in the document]
jlurien: As we talked yesterday, we should treat transversal aspects in the other forums (via Commonalities) and discussion in issue #105 specifically. Regarding what CAMARA lacks:

1. CAMARA does not exclude client_credentials at all, indeed we are using client_credentials in other CAMARA APIs so far. Each API should include as part of the yaml the securitySchemes to be considered.
2. Agree that this is still open and needs to be closed, but we expect that whatever is agreed at CAMARA level apply to all APIs under CAMARA umbrella. And it has not to be the same that is already in place for Mobile Connect, even if there are of course many aspects in common (OIDC, etc). In the end we hope that all CAMARA APIs share a similar look&feel to devs. If we as Telcos decided to go with Mobile Connect approach, probably we wouldn't need to copy that work here, and just reference the existing recommendations.
3. I don't see here a "recommended" approach but a different approach. Having dedicated endpoints per specific use case has many advantages in my opinion, as they are more focused on the specific requirements of the use case, for example you can enrich the request with parameters that are specific to that use case (as in some POST requests or GETs with query parameters), response is simpler and does not return information which is not required by the use case.

With controversial I'm just raising that we have a disagreement on whether CAMARA APIs should tackle this use case in the same way than Mobile Connect does. But I'm not questioning the work done within Mobile Connect, or implying that there is anything bad in MC.

documentation/API_documentation/API_definition.md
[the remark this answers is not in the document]
jlurien: This is not exactly the same functionality. A dedicated max_age parameter allows the consumer to decide on the threshold, within the limits that the API allows, and this limits may be global or controlled by 400 errors per implementation. It would be similar to the accuracy parameter that we have in Device Location API, where the consumer also pass the desired threshold within some limits.

In the end, it depends on which use case we want to provide. This is a good example of what I commented before about the advantage of having dedicated endpoints. If needed, we could easily define an specific parameter here, while it would be weird to pass this information to a generic /userinfo endpoint.

PR #6, opened 2022-12-13
simSwap API yaml

bigludo7:
@DT-DawidWroblewski 
Following our exchange on this topic, I was expecting to get a swagger close to this one.
**This is absolutely not a final contribution** but here we have both flavor: endpoint to accommodate MC implementation and simple endpoint.
Are we still aligned on this specification 'pattern'?

### DIFF
--- a/code/API_definitions/checkSimSwap-v0.4.0.camara.swagger.yaml
+++ b/code/API_definitions/checkSimSwap-v0.4.0.camara.swagger.yaml
@@ -1,0 +1,336 @@
+openapi: 3.0.3
+info:
+  title: Check simSwap
+  description: |+
+    Check simSwap API provides the customer the ability to obtain information on any recent SIM pairing change
+     related to the User’s mobile account. 
+
+    The API provided 3 operations:
+
+    - POST retrieveSimSwapDate : provide timestamp of latest sim swap,
+    - POST checkSimSwap: check if sim swap has been performed during a past period.
+    - POST openid/userInfo : retrieve information as defined in Account Takeover Protection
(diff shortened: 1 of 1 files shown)


code/API_definitions/checkSimSwap-v0.4.0.camara.swagger.yaml
```diff
@@ -0,0 +1,336 @@
...
+              schema:
+                $ref: "#/components/schemas/responseATP"
+components:
+  securitySchemes:
+    oAuth2ClientCredentials:
+      type: oauth2
+      flows:
+        clientCredentials:
+          tokenUrl: '{tokenUrl}'
+          scopes: {}
```
DT-DawidWroblewski, line 177: for MC ATP, we're using scope=mc_atp
"grant_type" should passed in query parameters, like this:
>/token?grant_type=client_credentials&scope=mc_atp

code/API_definitions/checkSimSwap-v0.4.0.camara.swagger.yaml
```diff
@@ -0,0 +1,336 @@
...
+        account_state:
+          type: string
+          enum:
+            - active
+            - inactive
+          example: active
+          description: >-
+            Status of the Mobile Connect account
+      required:
+        - sub
```
DT-DawidWroblewski, line 241: sub is optional

bigludo7:
Following Dec13th, 2022 meeting, the team prefer to have 2 distinct specifications defined instead of one. This PR is abandoned.

PR #8, opened 2022-12-14
adding MC ATP yaml

DT-DawidWroblewski:


patrice-conil (changes requested):
This openapi definition is incorrect, you need to check semantic and syntax erors using https://editor.swagger.io/.
For axample in schemas you need to group "required" attribute in a separate section.
```
    atpTimestamp:
      type: object
      properties:
        simChange:
          type: string
          required: true
          example: 'simChange: 2022-12-06'
        isUncontidionalCallDivertActive:
          type: string
          required: false
```
should be

```
    atpTimestamp:
      type: object
      properties:
        simChange:
          type: string
          example: 'simChange: 2022-12-06'
        isUncontidionalCallDivertActive:
          type: string
        required: 
          - simChange
```
I can help to fix these issues.

code/API_definitions/MC_ATP.yaml
```diff
@@ -0,0 +1,160 @@
...
+          type: boolean
+        deviceChange:
+          type: string
+        accountState:
+          type: string
+          enum:
+            - active
+            - inactive
+      required:
+        - simSwap
```
jlurien, line 141: this property is not listed above

code/API_definitions/MC_ATP.yaml
```diff
@@ -0,0 +1,160 @@
...
+                token_type: Bearer
+                expires_in: '500'
+  /openid/userinfo:
+    post:
+      tags:
+        - Mobile Connect ATP
+      summary: User Info
+      description: userinfo
+      requestBody:
+        content: {}
```
jlurien, line 64: No body definition?

jlurien (commented):
I made a quick review with some comments for your consideration. 
You can check also these errors at https://editor.swagger.io/#/:

Semantic error at security.0
Security requirements must match a security definition
Jump to line 118
Semantic error at security.1
Security requirements must match a security definition
Jump to line 119

DT-DawidWroblewski:
Yep - I saw this and fixing the swagger.
________________________________

patrice-conil (approved):

DT-DawidWroblewski:
@bigludo7 can you please review the code today and approve? thx!

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,201 @@
...
+|Parameter|Usage Category|Description|
+|:---|:---:|:---|
+|grant_type|REQUIRED|Value MUST be set to “client_credentials”.|
+|scope|REQUIRED|The scope of the access request as described in Section 3.3 of the OAuth2.0 specification. For ATP scope value is: **"mc_atp"**|
+
+**Table 1 - Mobile Connect ATP Client Credentials request parameters**
+
+### Resource Request
+
+Folllwing mechanism is defined to indicate the MC User whose data is being queried in the request. Namely, the following HTTP headers MUST be used for that purpose:
```
bigludo7, line 78: Following not Folllwing

documentation/API_documentation/API_definition.md
```diff
@@ -0,0 +1,201 @@
...
+|HTTP Header|Usage Category|Description|
+|:---|:---|:---|
+|User-ID-Type|REQUIRED [if the bearer type access token provided in the request is generic]|MC User identification type used in the User-ID header. One of these values MUST be used: <ul><li>**MSISDN**: Indicates the User-ID contains a plain MSISDN in international format according to ITU-T recommendation E.164. The plus sign (+) MUST NOT be included as a prefix.</li><li> **ENCR_MSISDN**: Indicates the User-ID contains an encrypted MSISDN. Support for the MSISDN user ID type is REQUIRED, whereas for ENCR_MSISDN is OPTIONAL.</li><ul>|
+|User-ID|REQUIRED [if the bearer type access token provided in the request is generic]|MC User identification value as per the type indicated in the HTTP header above.|
+
+Both the names and the values of these HTTP headers MUST be treated as case insensitive.
+
+### ATP API details
+
+YAML proposal:
```
bigludo7, line 89: Not sure we should keep this paragraph. Could be confusing for reader.

code/API_definitions/MC_ATP.yaml
```diff
@@ -0,0 +1,160 @@
+openapi: 3.0.0
+info:
+  title: CAMARA simSwap
```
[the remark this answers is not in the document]
bigludo7: We need a global consistency for this, as for numberVerify API the title is CAMARA numberVerify. We need a common policy for all API to decide if CAMARA is part of the title or not.

code/API_definitions/MC_ATP.yaml
```diff
@@ -0,0 +1,160 @@
...
+  description: CAMARA simSwap API based on Mobile Connect Account Takeover Protection API definition.
+  version: 0.1.0
+servers:
+  - url: https://api.server.test
+    description: API server providing CAMARA APIs
+tags:
+  - name: Mobile Connect ATP
+paths:
+  /token:
```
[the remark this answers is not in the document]
bigludo7: In CAMARA numberVerify proposal we have endpoints for /authorize & /token. We should have same pattern everywhere.
As this proposal is based on MC I'm fine with these endpoints but probably we could add a comment to indicate that they could be in a distinct server as mentionned by @jlurien 

BTW, @jlurien  in QoD API we have in the same swagger for both /sessions & notifications and they will not be implemented in same server.

jlurien:
@bigludo7  Regarding "in QoD API we have in the same swagger for both /sessions & notifications and they will not be implemented in same server."

indeed it is a different server. In that case we opted to add a disclaimer in description:

> ```
>       description: |
>         Important: this endpoint is to be implemented by the API consumer.
>         The QoD server will call this endpoint whenever any network related event occurs.
>         Currently only SESSION_TERMINATED event is implemented. Any other network events are ignored.
> ```

code/API_definitions/MC_ATP.yaml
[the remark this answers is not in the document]
DT-DawidWroblewski: Hi! Sure - I agree. Question - where we should make differentiation - on a folder level? Therefore, in API description folder we should have OAS and MC folders with separated simSwap yaml files?

code/API_definitions/MC_ATP.yaml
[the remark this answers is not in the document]
DT-DawidWroblewski: We should leave this comment inside swagger or in MD file?

code/API_definitions/MC_ATP.yaml
DT-DawidWroblewski: outdated

documentation/API_documentation/API_definition.md
DT-DawidWroblewski: fixing...

documentation/API_documentation/API_definition.md
DT-DawidWroblewski: Let's talk about it. With this paragraph, we give the reader direct links to yaml files. Let's change name, and instead of DT & Orange have MC and OAS.

code/API_definitions/MC_ATP.yaml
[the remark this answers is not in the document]
jlurien: I would document it in both. Problems with yaml may rise when some client library try to genere code directly from yaml definition. We should be more careful with these aspects across all APIs when final v1 versions are released.

documentation/API_documentation/API_definition.md
bigludo7: Yes this is better with MC & OAS.

DT-DawidWroblewski: done...

bigludo7 (approved):
Look good for me as a first version.
Agreed with @jlurien about /authorize & /token endpoints but should tackle this question globally for all API. This is probably a point worth to be discussed  in Commonalities --> we need an issue for that.

PR #37, opened 2023-05-23
POST for openid/userinfo is not compatible with OIDC

### LINKED ISSUE #27: POST for openid/userinfo is not compatible with OIDC
https://github.com/camaraproject/SimSwap/blob/c317b025fe1fd121cc25219ab9244471fe234c9d/code/API_definitions/MobileConnect/MC_ATP.yaml#L75

Correct me if I'm wrong, but IDY.24 implements OIDC openid/userinfo API, not juts coincidentally uses the same endpoint.
OIDC describe this endpoint as a GET API. Using POST here is incompatible with OIDC and breaks REST design principles ("POST UserInfo" looks like an update/create API).

DT-DawidWroblewski:
Fixes #27
+MoMs


monamok:
Hello!
@DT-DawidWroblewski I don't quite get the changes that you're making in this PR.

Why are you changing old MoMs from 2022? 

I think the problem is that the PR goes against MoM branch, not the master branch so there are undesired changes included.
Also my suggestion is not to include changes in the API definition and MoMs in the same PR.

PR #52, opened 2023-09-06
Consolidate api information into yaml file

### LINKED ISSUE #42: Embedding documentation in CAMARA API spec and decision about documenting authorization methods
**Problem description**

API documentation should be included in API Spec.
Related to: https://github.com/camaraproject/WorkingGroups/issues/164

**Expected action**
We must move the information from here https://github.com/camaraproject/SimSwap/blob/main/documentation/API_documentation/CAMARA/Check_sim_swap_API.md into the API yaml file to be aligned with the guidelines

**Additional context**

Checking the documentation as well as the API spec, for authorization it says that the API considers 2 security models: Three-legged access tokens with dedicated scopes, and OAuth 2.0 client credentials.

@bigludo7 having in mind that this API handles personal information how do you feel about removing OAUTH 2.0 client credentials grant and sticking to 3-legged tokens?

The documentation and API Spec must get updated by this decision.

fernandopradocabrillo:
#### What type of PR is this?

* cleanup
* documentation

1. Include API documentation into yaml file and remove extra md files
2. Rename api spec file to "sim_swap.yaml" since previous name "checkSimSwap.camara.swagger.yaml" no longer reflects the actual status of the API:
    * `check` is no longer the only operation available in the spec
    * there is no need to keep the distinction of `camara` since we no longer have MC version
    * since we are developing in openapi I think we don't need to include the `swagger` reference

Fixes (https://github.com/camaraproject/SimSwap/issues/42)

Issue #42 still has an open issue regarding the authorization flows to be implemented in the API, as we were waiting for a resolution in the discussions on Identity and Consent. This decission could lead to changes in the requests properties [(Link to Mona's comment)](https://github.com/camaraproject/SimSwap/issues/42#issuecomment-1604302363) so, if you are okey with it @bigludo7, @DT-DawidWroblewski, I suggest moving that topic to a new Issue and close #42 with the embeded documentation and the complete cleanup. 

#### Changelog input

N/A

#### Additional documentation 

N/A


code/API_definitions/sim_swap.yaml
```diff
@@ -2,15 +2,39 @@ openapi: 3.0.3
 info:
   title: SIM Swap
   description: |
-    SIM Swap API provides the customer the ability to obtain information on any recent SIM pairing change related to the User's mobile account.
+    The SIM swap API provides a programmable interface for developers and other users (capabilities consumers) to request the last date of a SIM swap performed on the mobile line, or, to check whether a SIM swap has been performed during a past period.
     
-    [GSMA Mobile Connect Account Takeover Protection specification](https://www.gsma.com/identity/wp-content/uploads/2022/12/IDY.24-Mobile-Connect-Account-Takeover-Protection-Definition-and-Technical-Requirements-v2.0.pdf) was used as source of input for this API. For more about Mobile Connect, please see [Mobile Connect website](https://mobileconnect.io/).
```
DT-DawidWroblewski, line 7: @fernandopradocabrillo I think we agreed to leave this reference in documentation... please change accordingly and do not delete this piece.

code/API_definitions/sim_swap.yaml
```diff
@@ -35,13 +59,14 @@ paths:
   /retrieve-date:
     post:
       security:
-        - oAuth2ClientCredentials: []
```
DT-DawidWroblewski, line 38: what about countries where oauth2ClientCredentials is allowed? we should keep this.

code/API_definitions/sim_swap.yaml
```diff
@@ -76,13 +104,14 @@ paths:
   /check:
     post:
       security:
-        - oAuth2ClientCredentials: []
```
DT-DawidWroblewski, line 79: what about countries where oauth2ClientCredentials is allowed? we should keep this.

code/API_definitions/sim_swap.yaml
```diff
@@ -116,21 +148,9 @@ paths:
           $ref: "#/components/responses/Generic504"
 components:
   securitySchemes:
-    oAuth2ClientCredentials:
```
DT-DawidWroblewski, line 119: we cannot delete this option as oauth2clientCredentials is an option that is available on many markets

code/API_definitions/sim_swap.yaml
fernandopradocabrillo: yes @DT-DawidWroblewski It is still there, I moved it to the section Further info and support.

code/API_definitions/sim_swap.yaml
fernandopradocabrillo: The idea that is being discuss in the commonalities Issue (https://github.com/camaraproject/Commonalities/issues/53) is to encapsulate the Authorization flows into the .well-known file of the OIDC protocol. This does not mean that we removed the client_credentials grant type, it is just not explicitly described inside the securitySchemes

code/API_definitions/sim_swap.yaml
fernandopradocabrillo: Idem

code/API_definitions/sim_swap.yaml
fernandopradocabrillo: Idem

code/API_definitions/sim_swap.yaml
DT-DawidWroblewski: I'd keep it to avoid confusion while reading the yaml file

code/API_definitions/sim_swap.yaml
DT-DawidWroblewski: idem

code/API_definitions/sim_swap.yaml
DT-DawidWroblewski: idem

code/API_definitions/sim_swap.yaml
bigludo7: @DT-DawidWroblewski may I suggest you also raise your concern  in the https://github.com/camaraproject/Commonalities/issues/53 ? i think it is important to provide your feedback at this level as it will impact all API.
@fernandopradocabrillo probably better to wait Commonalities outcome on this PR before to apply no ?

fernandopradocabrillo: @bigludo7 yes I also think we can wait until the commonalitoes PR is merged. I was just advancing some work!

hdamker:
@fernandopradocabrillo 

I strongly propose to have two different PRs:
* One to consolidate the documentation into the YAML file. It's documentation only and will not impact the upcoming version number of a release (or just the patch number)
* Changes of the YAML file, e.g. discussion about security schema

Doing it in one PR will make it complex and will block the independent changes.

fernandopradocabrillo:
> @fernandopradocabrillo
> 
> I strongly propose to have two different PRs:
> 
> * One to consolidate the documentation into the YAML file. It's documentation only and will not impact the upcoming version number of a release (or just the patch number)
> * Changes of the YAML file, e.g. discussion about security schema
> 
> Doing it in one PR will make it complex and will block the independent changes.

@hdamker @bigludo7 @DT-DawidWroblewski Updated this PR to only include the documentation in the yaml. I'll create a new one with then changes in the scopes, headers, etc

DT-DawidWroblewski:
it is ok. thanks @fernandopradocabrillo

fernandopradocabrillo:
> Look good for me. I'm fine for this one to keep same version as we just added documentation but probably in future update we should change the last didit version no?

Yes, I have updated to v0.5.0-wip in the PR with the yaml changes. I don't know exactly what convention are we following in the versioning so I'm open to suggestions!

Issue #49, opened 2023-08-29
Add x-correlator as header attribute


bigludo7:
**Problem description**
As defined in the Commonalities - Design guideline, X-Correlator attribute in header must be added

**Possible evolution**
Add this attribute.

**Alternative solution**

**Additional context**

hdamker:
@bigludo7 

I would like to understand better where the CAMARA design guidelines require to add this header attribute into the OAS file. I don't question that the header attribute must be supported if the API caller sends it, I'm just questioning that it must be included in the OAS specification and clutter it (there are lot of well-known headers which need to be supported). I'm especially asking as your issue raised here is for me a precdent for other sub project.

In more detail:  chapter "[3.5 HTTP Header Definitions](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#35-http-headers-definition)" list a lot of HTTP headers and says rightfully: 

> To avoid cluttering the CAMARA OAS (Swagger) definition files, the above headers must not be included explicitly in the API definition files even when supported, as it can be assumed that developers will already be familiar with their use.

Then there is chapter "9. Architecture Headers" which is unfornately not linked with 3.5. It has a table of  "common headers" which "should be included". One ot the two is `X-Correlator`. But first it isn't a "must", the API Caller isn't required to send it, and - most relevant for the issue here - the table says exlicitly that it is not "required within the OAS file".

If you agree with my interpretation, we should close the issue here and open one in Commonalities to make the Guidelines less ambiguous, e.g. list "X-Correlator" also in chapter 3.5 as a common header and add some explanation withinn chapter 9 what the chapter means for API implementations.

bigludo7:
@hdamker 
I've raised this issue initially here following feedback from developer implementing our API who did not understand why Number Verification API  feature the `X-Correlator` while not SimSwap API. As they implemented both API the inconsistency was highlighted. I was chasing a quick resolution on this but probably better to have also the global picture of all API

Agreed with you on the way to proceed: 

- We close this issue
- We need to update the design guideline. I'm not sure about this chapter 9 value. As you write, we should move the information in this chapter 9 in a paragraph of 3.5.
- Probably for global consistency we should enforce presence of `x-correlator`  in all our api.
- Then we can again raise issue in API without this header parameter.

bigludo7:
@eric-murray About Design Guidelines as 'source of truth' completely aligned- if we stated it that it is an UUID we must have it as UUID in the API.

[...]

PR #53, opened 2023-09-07
Include OIDC securityScheme, format scopes and add x-correlator header

fernandopradocabrillo:
#### What type of PR is this?

* enhancement/feature

1. Align securitySchemes with the current commonality proposal to use OpenId to encapsulate grant types and authorization flows
2. Change the scopes format to the current proposal of commonalities and align with the rest of CAMARA APIs
3. Include X-Correlator headers according to [Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#9-architecture-headers)
4. Update API version to 0.5.0-wip

Sources for points 1 and 2:
    * Commonalities Issue: https://github.com/camaraproject/Commonalities/issues/53
    * Identity and Consent Issue: https://github.com/camaraproject/IdentityAndConsentManagement/issues/57
    * Commonalities PR: https://github.com/camaraproject/Commonalities/pull/57

Fixes (https://github.com/camaraproject/SimSwap/issues/49) Add x-correlator as header attribute

Notice that the documentation is still not update, once we merge the documentation PR (https://github.com/camaraproject/SimSwap/pull/52) I'll sync this one

#### Changelog input

N/A

#### Additional documentation 

N/A


bigludo7 (approved):
LGTM/

fernandopradocabrillo:
@bigludo7 I updated the PR after the merge of the documentation PR. Now it is synchronized with the documentation

bigludo7 (dismissed):
LGTM/

code/API_definitions/sim_swap.yaml
```diff
@@ -59,13 +59,14 @@ paths:
   /retrieve-date:
     post:
       security:
-        - oAuth2ClientCredentials: []
```
DT-DawidWroblewski, line 62: If we delete this option, then we can have a clear message to developer that the ONLY way to consume this API is to use OpenID Connect - so no client credential option is available.

code/API_definitions/sim_swap.yaml
```diff
@@ -140,21 +148,21 @@ paths:
           $ref: "#/components/responses/Generic504"
 components:
   securitySchemes:
-    oAuth2ClientCredentials:
```
DT-DawidWroblewski, line 143: idem

code/API_definitions/sim_swap.yaml
DT-DawidWroblewski: not really... 
[OpenID Connect overview](https://openid.net/specs/openid-connect-core-1_0.html#Overview)

fernandopradocabrillo: @DT-DawidWroblewski I think this would be best discussed in the Identity and Consent WG, as that is where the change in authentication methods was approved.

hdamker (changes requested):
To have a base release within this repository, I propose to create an (alpha) release v0.4.1 of the current content in main before this PR will be merged and you continue with v0.5.0-wip. This also as some companies are want to start implemantation on a release version, knowing that they have to adapt the security schema and other points later.

hdamker:
In addition the same comments as in https://github.com/camaraproject/NumberVerification/pull/60#pullrequestreview-1629302593 ... in my view X-Correlator does not need to be defined within the OAS here.

fernandopradocabrillo:
> @fernandopradocabrillo
> 
> Unfortunately this PR is addressing two different issues which have unrelated discussions. I would recommend to reduce the scope of this PR to securityScheme and format scopes, but carve the X-Correlator discussion out into a separate PR later. That will allow that the discussion of in #49 will not block the PR (and in consequence a first release).

@hdamker Sure, I will remove the x-correlator from this PR and open a new one so we only have the securitySchemes and scopes changes here

eric-murray:
I might be missing something here, but why does this version of the API still require the MSISDN (phone number) to be passed in the `/check` and `/retrieve-date` calls when this will already be known from the authorisation token? I can see that it was required when client credentials were an option, but don't understand why it is still required for the three-legged flow as user consent is checked in a previous step. Maybe someone could enlighten me.

fernandopradocabrillo:
> I might be missing something here, but why does this version of the API still require the MSISDN (phone number) to be passed in the `/check` and `/retrieve-date` calls when this will already be known from the authorisation token? I can see that it was required when client credentials were an option, but don't understand why it is still required for the three-legged flow as user consent is checked in a previous step. Maybe someone could enlighten me.

@eric-murray I also think it should be removed. My idea was to first solve the authorization topic and the create an issue and a PR to remove it, since the new version is still going to be on wip until we go for a release. Step by step 😄

eric-murray:
Thanks @fernandopradocabrillo for the prompt response. So I understand that removing this parameter will be discussed in a later step. So now I'm no longer confused! Many thanks.

gregory1g:
> @eric-murray I also think it should be removed. My idea was to first solve the authorization topic and the create an issue and a PR to remove it, since the new version is still going to be on wip until we go for a release. Step by step 😄

This was discussed earlier, when both client credentials and 3-legged where considered. Simply to keep payload auth independent. However, if 3-legged will be the only auth method, MSISDN in the input becomes redundant.

MSISDN is declared as a sensitive information, this leads to special handling (like putting it into header to reduce risk of occasional leakage).   From this point of view, removing MSISDN from the payload completely is the best protection.

BUT, we should not postpone this step for a long time - removing MSISDN is a not backward compatible change.

fernandopradocabrillo:
@gregory1g @eric-murray 
Issue and PR created to solve this issue:
Issue: https://github.com/camaraproject/SimSwap/issues/58
PR: https://github.com/camaraproject/SimSwap/pull/59

fernandopradocabrillo:
> > @fernandopradocabrillo
> > Unfortunately this PR is addressing two different issues which have unrelated discussions. I would recommend to reduce the scope of this PR to securityScheme and format scopes, but carve the X-Correlator discussion out into a separate PR later. That will allow that the discussion of in #49 will not block the PR (and in consequence a first release).
> 
> @hdamker Sure, I will remove the x-correlator from this PR and open a new one so we only have the securitySchemes and scopes changes here

@hdamker 
Update the PR removing the x-correlator changes

eric-murray:
@DT-DawidWroblewski 
The decision is now to retain Client Credentials as a valid authentication scheme for SIM Swap? As your meetings are not minuted, it can be difficult to follow.

fernandopradocabrillo:
> @DT-DawidWroblewski The decision is now to retain Client Credentials as a valid authorisation scheme for SIM Swap? As your meetings are not minuted, it can be difficult to follow.

@eric-murray 
from Telefonica side, as stated in the meeting, we don't agree with that. We are okey maintaining the securitySchemes as OpenId which is what is being agreed in commonalities.
We should merge the PR with the agreed changes and discuss in commonalities if there should be an indication of the allowed auth flows inside the yaml.

This PR seems to be going in that direction: https://github.com/camaraproject/Commonalities/pull/57

eric-murray:
Thanks @fernandopradocabrillo. In the absence of meeting minutes, I shall wait to see if this PR is approved and merged in its current state.

DT-DawidWroblewski:
Hi @eric-murray !

you can find MoMs in dedicated branch, here:

[MoM](https://github.com/camaraproject/SimSwap/tree/MoM/documentation/MeetingMinutes)

I was asked to make a clean-up and add MoMs to main branch, so I'll do it later this week.

code/API_definitions/sim_swap.yaml
```diff
@@ -42,7 +49,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.4.1
+  version: 0.8.0
```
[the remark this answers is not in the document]
DT-DawidWroblewski: got it - as long as we can converge towards release, version is not an issue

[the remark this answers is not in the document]
DT-DawidWroblewski: we stick to 0.5.0-wip as we're changing securityScheme based on consensus we reached on last meeting

DT-DawidWroblewski:
@bigludo7 can you please review changes, so we can merge them and create a release version? thx!

code/API_definitions/sim_swap.yaml
```diff
@@ -30,6 +30,13 @@ info:
 
     - POST check: Checks if SIM swap has been performed during a past period (defined in the request with 'maxAge' attribute) for a given phone number.
 
+    The API supports 2 security security schemes:
+
+    - clientCredentials
```
fernandopradocabrillo, line 35: @bigludo7 @DT-DawidWroblewski @eric-murray 
On this topic (at the risk of sounding repetitive) there is no agreement yet.. 
I understand the need to provide a release version, which I completely agree with, but it is a complicated issue that involves conversations in other WG at different levels.
In my opinion, we should release a version with the securitySchemes agreed (OpenId) and open a new PR to provide more information in the yaml description if that is decission on commonalities.

DT-DawidWroblewski: Hi @fernandopradocabrillo !

We agreed on our last sub-group meeting to have openIdConnect security scheme with well-known, where developer derives available security schemes. Additionaly, we provide information about available security schemes in description.

Alternatively, we can leave our agreement and simply create a release version based on existing API description that explicitly covers clientCredentials & three-legged-token, but that is against our consensus.

I suggest to continue with this PR and open a new one, that observes the outcome from other WGs.

Otherwise, we cannot converge towards release version...

fernandopradocabrillo: Hi @DT-DawidWroblewski !
My comment is mainly focused on the main issue we are dealing with here. We cannot explicitly say that we support customer credentials because that is not what has been agreed. And this is a discussion that unfortunately we cannot have in the subgroup and decide unilaterally, it has to be agreed in I&C and Commonalities.
In order to proceed with the release, including the OpenId securityScheme allows us to unlock this issue by delegating this decision to the I&C discussion and revisit it in the future if there is a new agreement.

DT-DawidWroblewski: We agreed that we want to give developer a hint what security schemes are available for SimSwap API, if a particular Auth Server has such capabilities. Can you suggest a place in yaml where we can add this? Thx!

bigludo7: @DT-DawidWroblewski @fernandopradocabrillo I'm bit lost now.

> Can you suggest a place in yaml where we can add this?

If we use open id as described for example here https://swagger.io/docs/specification/authentication/openid-connect-discovery/ the objective is to **not** mix authorization and API yaml definition.

But @DT-DawidWroblewski I understood that it will **not** prevent you to use client credential as long as it is defined in the https://server.com/.well-known/openid-configuration file.

If we have in the yaml both openId & client credential listed that will be confusing no ?

DT-DawidWroblewski: I saw a comment inside [Id&C working group](https://github.com/camaraproject/IdentityAndConsentManagement/issues/57#issuecomment-1755365845). In order to be compliant with OAS, we should explicitly state both security schemes.

If we don't want to follow the consensus we made during a meeting, I suggest waiting until Issue #57 inside Id&C working group is closed.

DT-DawidWroblewski:
Hi @fernandopradocabrillo @bigludo7  !

I updated API description according to recent OGW suggestion.

Please review/approve PR, so we can merge it and create release version.

Thx!
David

bigludo7:
@fernandopradocabrillo @DT-DawidWroblewski 
Same than @fernandopradocabrillo - As we have already the 0.4.1 should be 0.5.0-wip or 0.4.2-wip 
thanks

DT-DawidWroblewski:
Ok - I think that v0.5.0 makes sense :) 

I'll update accordingly...

DT-DawidWroblewski:
@fernandopradocabrillo  @bigludo7  please review

code/API_definitions/sim_swap.yaml
```diff
@@ -140,21 +138,9 @@ paths:
...
-      flows:
-        authorizationCode:
-          authorizationUrl: https://auth.example.com/authorize
-          tokenUrl: https://auth.example.com/token
-          scopes:
-            check-sim-swap: checkSimSwap operation
-            retrieve-sim-swap-date: retrieveSimSwapDate operation
+    openId:
+      type: openIdConnect
+      openIdConnectUrl: https://example.com/.well-known/openid-configuration
```
eric-murray, line 143: I know this is currently a WIP but, given the existence of extension grants, the `.well-known/openid-configuration` could include any grant type that has a defined `urn`. Could we not give some guidance to those looking to implement this via a `description` documenting what grant types they might find in `openid-configuration` for this API and which they won't?

Or is that just too controversial at the moment?

fernandopradocabrillo: For the it's been agreed to not indicate further information inside the yaml. I think there are further details regarding this in the I&C workgroup

fernandopradocabrillo:
> @fernandopradocabrillo @bigludo7 please review

LGTM

I can approve because it's my PR but it's fine from my side :)

code/API_definitions/sim_swap.yaml
eric-murray: Can you point me to the **agreement** that says YAMLs **must not** say anything about what security schemes are applicable for that API?

fernandopradocabrillo: I'm checking with my collegues and it seems I wasn't align with the final decisions on this topic and all the details. I'm sorry for the misunderstanding. From TEF we are align with what was decided in the TSC.
Would it be okey to merge this PR with the schemes and continue with the discussion of appropiate wording/way to put this information in a different issue? 
It seems that this is something that affects all apis and we should document in the same way for all apis.

bigludo7 (approved):
LGTM

code/API_definitions/sim_swap.yaml
eric-murray: OK, let's return to this topic later. It needs to be resolved before any release.

DT-DawidWroblewski (approved):
LGTM

code/API_definitions/sim_swap.yaml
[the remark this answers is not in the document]
fernandopradocabrillo: @hdamker 
This comment has been addressed, can you remove the change request to merge the PR? Thanks!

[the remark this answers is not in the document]
hdamker: @fernandopradocabrillo yes, this is addressed.

hdamker (approved):
Approval to remove my change request which was not related to topic at hand. 

But there is the need to revist the documentation towards developers before a release of the API as stated multiple times above in comments, latest by @eric-murray in https://github.com/camaraproject/SimSwap/pull/53/#discussion_r1374511984

Issue #58, opened 2023-09-25
Remove MSISDN from request body


fernandopradocabrillo:
**Problem description**
With the use of 3-legged authorization flow, it is no longer necessary to send the phone number in the request body as it can be obtained from the token.

**Expected behavior**
The MSISDN will be obtained from the 3-legged token.

**Alternative solution**
N/A
**Additional context**
N/A

DT-DawidWroblewski:
Hi @fernandopradocabrillo !

as client credentials is still an option MSISDN has to be passed in a resource request. However, I suggest using HTTP request header to pass the MSISDN value.

chrishowell:
I don't think we can remove the `msisdn` from request, there could very easily be scenarios where the target device (the `msisdn` in the request) is not the same as the authorisation device.

fernandopradocabrillo:
Hi @chrishowell 
In today's meeting we have agreed to leave it as an optional parameter in the request body.
I didn't have the time to update the PR yet

PR #59, opened 2023-09-25
Remove msisdn from request body

fernandopradocabrillo:
#### What type of PR is this?

* correction

Remove phone number (MSISDN) from request body

Fixes (https://github.com/camaraproject/SimSwap/issues/58)

There are other PRs with changes that will be applied together. Once previous PRs are merged I will update and align this one.

#### Changelog input

N/A

#### Additional documentation 

N/A

### DIFF
--- a/code/API_definitions/sim_swap.yaml
+++ b/code/API_definitions/sim_swap.yaml
@@ -66,6 +66,14 @@ paths:
         - Retrieve SIM swap date
       description: Get timestamp of last MSISDN <-> IMSI pairing change for a mobile user account provided with MSIDN.
       operationId: retrieveSimSwapDate
+      requestBody:
+        description: |
+          Create a SIM swap date request for a MSISDN identifier.
+        content :
+          application/json:
+            schema:
+              $ref: "#/components/schemas/CreateSimSwapDate"
+        required: true
       responses:
         "200":
           description: Contains information about SIM swap change
@@ -162,6 +170,11 @@ components:
           type: boolean
           description: Indicates whether the SIM card has been swapped during the
             period within the provided age.
+    PhoneNumber:
+      type: string
+      pattern: '^\+?[0-9]{5,15}$'
+      example: '+346661113334'
+      description: Subscriber number in E.164 format (starting with country code). Optionally prefixed with '+'.
     ErrorInfo:
       type: object
       required:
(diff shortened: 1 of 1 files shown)


bigludo7:
Hello @fernandopradocabrillo 
Sorry if my question is stupid but if we remove the msisdn the API will be unuseable in Wifi or Tetheering situation right ?

fernandopradocabrillo:
Hi @bigludo7
So, AFAIK, I don't see the relation with the wifi or tethering case. If you are in the situation of using one of those you won't be able to use AuthCode since it uses network auth, so you will have to use CIBA passing the phone number as login_hint. This will associate the phone_number to the token so you won't have to pass the phone number to the endpoint.

gregory1g:
as far as I understand, when an application does not have an access token it will be redirected to OIDC endpoint, there it will send auth request together with MSISDN as login_hint (with needed scopes). OIDC server will check if this application has a permission to  requested scope and given MSISDN. If "yes" (pre-configured or granted by the user interactively - does not matter), OIDC server returns an access token (which is MSISDN specific). If no, and only in this case, OIDC could need to contact user to request their consent. 
So, if we assume that SimSwap is opt-in by default (based on a legal agreement between user and application provider), OIDC server will not need to contact a user. And if user explicitly opted-out, again, OIDC sees this and simply reject authorization without contacting the user.

DT-DawidWroblewski:
Based on current Local Champion activities in Germany it has been agreed to move forward with API release that contains client credentials security scheme. Therefore, it is required to have MSISDN passed inside resource request. Removing MSISDN makes sense where 3-legged-token is in place. Suggestion: keep phone_number as a option in a resource request (add proper description that refers to client credentials flow).

bigludo7:
@DT-DawidWroblewski the decision in Germany is only for the check endpoint or for both endpoints?

I'm wondering if we should not split this API in 2 for simplification.

- one yaml for the check endpoint and could be client credential as it seem that what the market has
- one yaml for the get simswap date and this one in three legged and probably not require the msisdn

gregory1g:
> Based on current Local Champion activities in Germany it has been agreed to move forward with API release that contains client credentials security scheme. Therefore, it is required to have MSISDN passed inside resource request. Removing MSISDN makes sense where 3-legged-token is in place. Suggestion: keep phone_number as a option in a resource request (add proper description that refers to client credentials flow).

So, we keep support for both options, right? If so, does it make sense to make MSISDN optional, so it can be skipped when 3-legged approach is used?

DT-DawidWroblewski (approved):
thx @fernandopradocabrillo !

all good - according to our last meeting

PR #60, opened 2023-10-02
Sim Swap Notification subscription yaml

### LINKED ISSUE #47: GSMA OGW Product WS - Subscription/notification mechanism for SIM SWAP formal request
Problem description
Formal inclusion of the GSMA OGW Product WS request to include Subscription/notification mechanism as part of OGW services.

https://github.com/camaraproject/Commonalities/issues/8#issuecomment-1642419272

Possible evolution
SIM SWAP API, as many other current APIs, may be enhanced by a subscription mechanism that will enable different use cases, e.g.:

- Antifraud mechanism may require being notified by the operator when a SIM is affected by a swap procedure, implementing different mechanisms to avoid fraud (e.g., warm or confirm with the user).

The proposal includes examples or proposal for input-outputs of such mechanism, to be discussed and aligned with the current service API.

Alternative solution
N/A

Additional context
More details included in the formal request file, attached
[Camara Submission - Subscription_Notifications v2.docx](https://github.com/camaraproject/SimSwap/files/12098247/Camara.Submission.-.Subscription_Notifications.v2.docx)

bigludo7:
#### What type of PR is this?

* enhancement/feature

Add a new API to manage sim swap notification subscription & notification.
yaml aligned with CAMARA commonalities design guidelines.

Fixes #47 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -215,19 +215,19 @@ paths:
                     code: INVALID_INPUT
                     status: 400
                     message: "Expected property is missing: subscriptionId"
-        401:
+        "401":
           $ref: "#/components/responses/Generic401"
-        403:
+        "403":
           $ref: "#/components/responses/Generic403"
-        404:
+        "404":
           $ref: "#/components/responses/Generic404"
-        500:
+        "500":
           description: Server error
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorInfo"
-        503:
+        "503":
           $ref: "#/components/responses/Generic503"
     delete:
       tags:
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -27,8 +27,8 @@ info:
       - ``org.camaraproject.sim-swap.v0.swapped`` - Event triggered when a sim swap occurs on the associated phoneNumber
 
 
-    Note: Additionnally to these list, ``org.camaraproject.sim-swa.v0.subscription-ends`` notification `type` is sent when the subscription ends. 
-    This notification did not requires dedicated subscription. 
+    Note: Additionnally to these list, ``org.camaraproject.sim-swap.v0.subscription-ends`` notification `type` is sent when the subscription ends. 
+    This notification does not require a dedicated subscription. 
     It is used when the subscription expire time (required by the requester) has been reached or if the API server has to terminate subscription (e.g. when the corresponding phone number is not served by the operator anymore). 
 
     ### Notification callback
@@ -52,7 +52,7 @@ info:
   version: 0.1.0-wip
 externalDocs:
   description: Product documentation at CAMARA
-  url: https://github.com/camaraproject/
+  url: hhttps://github.com/camaraproject/SimSwap
 servers:
   - url: '{apiRoot}/sim-swap-notification-subscriptions/v0'
     variables:
@@ -380,7 +380,7 @@ components:
       enum:
         - org.camaraproject.sim-swap.v0.swapped
 
-    RoamingNotificationEventType:
+    SimSwapNotificationEventType:
       type: string
       description: |
         swapped - Event triggered when a sim swap was performed for this phone number
@@ -449,9 +449,7 @@ components:
         source:
           $ref: '#/components/schemas/Source'
         type:
-          type: string
-          description: 'type of event as defined in each CAMARA API'
-          example: org.camaraproject.sim-swap.v0.swapped
+          $ref: '#/components/schemas/SimSwapNotificationEventType'
         specversion:
           type: string
           description: Version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version)
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -246,15 +246,15 @@ paths:
           schema:
             $ref: '#/components/schemas/SubscriptionId'
       responses:
-        204:
+        "204":
           description: event subscription deleted
-        202 :
+        "202":
           description: Request accepted to be processed. It applies for async deletion process.
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/SubscriptionAsync'
-        400:
+        "400":
           description: Invalid input
           content:
             application/json:
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -164,7 +164,8 @@ paths:
       description: Retrieve a list of sim swap event subscription(s)
       operationId: retrieveSubscriptionList
       security:
-        - openId::
+        - openId:
+
           - sim-swap:subscriptions:read
       responses:
         200:
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -73,7 +73,8 @@ servers:
         description: API root
       basePath:
         default: /sim-swap-notification-subscriptions/v0
-        description: Base path for the sim swap notification subscription API
+        description: Base path for the sim swap notification subscriptions API
+
 tags:
   - name: Sim swap notification subscription
     description: Operation to manage event subscription on sim swap event (swapped)
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -154,7 +154,7 @@ paths:
         - openId:
           - sim-swap:subscriptions:read
       responses:
-        200:
+        "200":
           description: List of event subscription details
           content:
             application/json:
@@ -162,15 +162,15 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/SubscriptionInfo'
-        400:
+        "400":
           $ref: '#/components/responses/Generic400'
-        401:
+        "401":
           $ref: '#/components/responses/Generic401'
-        403:
+        "403":
           $ref: '#/components/responses/Generic403'
-        500:
+        "500":
           $ref: '#/components/responses/Generic500'
-        503:
+        "503":
           $ref: '#/components/responses/Generic503'
   /subscriptions/{subscriptionId}:
     get:
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -164,7 +164,6 @@ paths:
       operationId: retrieveSubscriptionList
       security:
         - openId:
-
           - sim-swap:subscriptions:read
       responses:
         200:
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -324,7 +324,7 @@ components:
         - subscriptionDetail
       properties:
         subscriptionDetail:
-          $ref: '#/components/schemas/SubscriptionDetail'
+          $ref: '#/components/schemas/CreateSubscriptionDetail'
         subscriptionExpireTime:
           type: string
           format: date-time
@@ -350,6 +350,18 @@ components:
             OAuth2 token to be used by the callback API endpoint. It MUST be indicated within HTTP Authorization header e.g. Authorization: Bearer $notificationAuthToken
 
     SubscriptionDetail:
+      description: The detail of the requested event subscription
+      type: object
+      required:
+        - type
+        - phoneNumber
+      properties:
+        phoneNumber:
+          $ref: '#/components/schemas/PhoneNumber'
+        type:
+          $ref: '#/components/schemas/SimSwapEventType'
+
+    CreateSubscriptionDetail:
       description: The detail of the requested event subscription
       type: object
       required:
@@ -389,6 +401,15 @@ components:
           properties:
             subscriptionId:
               $ref: '#/components/schemas/SubscriptionId'
+            subscriptionDetail:
+              $ref: '#/components/schemas/SubscriptionDetail'
+            subscriptionExpireTime:
+              type: string
+              format: date-time
+              example: 2023-01-17T13:18:23.682Z
+              description: The subscription expiration time in date-time format.
+            webhook:
+              $ref: '#/components/schemas/Webhook'
             startsAt:
               type: string
               format: date-time
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -214,7 +214,7 @@ paths:
                   value:
                     code: INVALID_INPUT
                     status: 400
-                    message: "Expected property is missing: eventSubscriptionId"
+                    message: "Expected property is missing: subscriptionId"
         401:
           $ref: "#/components/responses/Generic401"
         403:
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -272,7 +272,7 @@ paths:
                   value:
                     code: INVALID_INPUT
                     status: 400
-                    message: "Expected property is missing: eventSubscriptionId"
+                    message: "Expected property is missing: subscriptionId"
         401:
           $ref: "#/components/responses/Generic401"
         403:
--- a/code/API_definitions/sim-swap-notification-subscription.yaml
+++ b/code/API_definitions/sim-swap-notification-subscription.yaml
@@ -1,0 +1,620 @@
+openapi: 3.0.3
+info:
+  title: Sim Swap Notification Subscription
+  description: |
+    This API provides the customer with the ability to subscribe to event related to a sim swap operation performed on an associated phone number.
+
+    # Introduction
+
+    # Relevant terms and definitions
+
+    * **SIM swap**:
+    A SIM swap is a process in which a mobile phone user's current SIM card is deactivated or replaced with a new one. This is typically done by contacting the user's mobile service provider and requesting a new SIM card for various reasons, such as a lost or damaged SIM card, upgrading to a new phone, or changing phone numbers while keeping the same device.
+
+    # API Functionality    
+
+    The API exposes following capabilities:
+
+
+    ## Sim Swap notification subscription
+
+    Theses endpoints allow to manage notification subscription on Sim Swap event. 
+    The CAMARA subscription model is detailed in the CAMARA API design guideline document and follows Cloudevents specification.
+
+    It is mandatory in the subscription to provide the event `type` subscribed are several are managed in this API.
+
+    Only one event ``type`` is managed for this API: 
+      - ``org.camaraproject.sim-swap.v0.swapped`` - Event triggered when a sim swap occurs on the associated phoneNumber
+
+
+    Note: Additionnally to these list, ``org.camaraproject.sim-swa.v0.subscription-ends`` notification `type` is sent when the subscription ends. 
+    This notification did not requires dedicated subscription. 
+    It is used when the subscription expire time (required by the requester) has been reached or if the API server has to stop sending notification prematurely. 
+
+    ### Notification callback
+
+    This endpoint describes the event notification received on subscription listener side when the event occurred. As for subscription, detailed description of the event notification is provided in the CAMARA API design guideline document.
+
+    **WARNING**: This callback endpoint must be exposed on the listener side as `POST /{$request.body#/webhook/notificationUrl}`
+      will be posted by the sim swap resource server to the webhook url provided by notification 
+      listener.
+
+    ## Further info and support
+
+    (FAQs will be added in a later version of the documentation)
+
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: project-email@sample.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.1.0-wip
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  #  - {}
+  - oAuth2ClientCredentials: []
+  #  - BasicAuth: []
+  #  - apiKey: []
+  - threeLegged:
+      - sim-swap:subscriptions:create
+servers:
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+      basePath:
+        default: /sim-swap-notification-subscriptions/v0
+        description: Base path for the sim swap notification subscription API
+tags:
+  - name: Sim swap notification subscription
+    description: Operation to manage event subscription on sim swap event (swapped)
+
+paths:
+  /subscriptions:
+    post:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Create a sim swap event subscription for a phone number'
+      description: Create a sim swap event subscription for a phone number
+      operationId: createSimSwapSubscription
+      security:
+        - threeLegged:
+          - sim-swap:subscriptions:create
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateSubscription'
+        required: true
+      callbacks:
+        notifications:
+          "{$request.body#/webhook/notificationUrl}":
+            post:
+              tags:
+                - Session notifications callback
+              summary: "Session notifications callback"
+              description: |
+                Important: this endpoint is to be implemented by the API consumer.
+                The sim swap server will call this endpoint whenever a swapped event occurs.
+              operationId: postNotification
+              requestBody:
+                required: true
+                content:
+                  application/json:
+                    schema:
+                      $ref: "#/components/schemas/CloudEvent"
+                    examples:
+                      SWAPPED:
+                        $ref: '#/components/examples/SWAPPED'
+                      SUBSCRIPTION_ENDS:
+                        $ref: '#/components/examples/SUBSCRIPTION_ENDS'  
+              responses:
+                "204":
+                  description: Successful notification
+                "400":
+                  $ref: "#/components/responses/Generic400"
+                "401":
+                  $ref: "#/components/responses/Generic401"
+                "403":
+                  $ref: "#/components/responses/Generic403"
+                "500":
+                  $ref: "#/components/responses/Generic500"
+                "503":
+                  $ref: "#/components/responses/Generic503"
+              security:
+                - { }
+                - notificationsBearerAuth: [ ]
+
+      responses:
+        201:
+          description: Created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SubscriptionInfo'
+        202 :
+          description: Request accepted to be processed. It applies for async creation process.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SubscriptionAsync'
+        400:
+          $ref: '#/components/responses/Generic400'
+        401:
+          $ref: '#/components/responses/Generic401'
+        403:
+          $ref: '#/components/responses/Generic403'
+        409:
+          $ref: '#/components/responses/Generic409'
+        500:
+          $ref: '#/components/responses/Generic500'
+        503:
+          $ref: '#/components/responses/Generic503'
+    get:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Retrieve a list of sim swap event subscription'
+      description: Retrieve a list of sim swap event subscription(s)
+      operationId: retrieveSubscriptionList
+      responses:
+        200:
+          description: List of event subscription details
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/SubscriptionInfo'
+        400:
+          $ref: '#/components/responses/Generic400'
+        401:
+          $ref: '#/components/responses/Generic401'
+        403:
+          $ref: '#/components/responses/Generic403'
+        500:
+          $ref: '#/components/responses/Generic500'
+        503:
+          $ref: '#/components/responses/Generic503'
+  /subscriptions/{subscriptionId}:
+    get:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Retrieve a sim swap event subscription for a phone number'
+      operationId: retrieveSubscription
+      description: retrieve event subscription information for a given subscription.
+      parameters:
+        - name: subscriptionId
+          in: path
+          description: Subscription identifier that was obtained from the create subscription operation
+          required: true
+          schema:
+            type: string
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SubscriptionInfo'
+        400:
+          description: Invalid input
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+              examples:
+                Generic400:
+                  summary: Schema validation failed
+                  value:
+                    code: INVALID_INPUT
+                    status: 400
+                    message: "Schema validation failed at  ..."
+                subscriptionIdRequired:
+                  summary: eventSubscription id is required
+                  value:
+                    code: INVALID_INPUT
+                    status: 400
+                    message: "Expected property is missing: eventSubscriptionId"
+        401:
+          $ref: "#/components/responses/Generic401"
+        403:
+          $ref: "#/components/responses/Generic403"
+        404:
+          $ref: "#/components/responses/Generic404"
+        500:
+          description: Server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+        503:
+          $ref: "#/components/responses/Generic503"
+    delete:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Delete a sim swap event subscription'
+      operationId: deleteSubscription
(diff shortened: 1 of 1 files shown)


code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,620 @@
...
+      operationId: retrieveSubscriptionList
+      responses:
+        200:
+          description: List of event subscription details
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/SubscriptionInfo'
```
fernandopradocabrillo, line 171: The same comment as in the creation body, the suscriptionDetails should be required in the responses of the GET operations.

The subscriptionId should be required too. Otherwise how would I be able to get the details of a particular subscription?

And, maybe the `startsAt` field should be required too. Since we have both models sync/async:
- If the response is async then the field is not present in the response
- If the response is sync, then it means that the subscription has already been created so it has a `startsAt` date and should be required since the information is available.

fernandopradocabrillo (commented):
@bigludo7 
I leave a few initial comments to discuss. 

Thanks for the PR!

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,620 @@
...
+    SubscriptionDetail:
+      description: The detail of the requested event subscription
+      type: object
+      required:
+        - type
+      properties:
+        phoneNumber:
+          $ref: '#/components/schemas/PhoneNumber'
+        type:
+          $ref: '#/components/schemas/SimSwapEventType'
```
gregory1g, line 381: should we refence RoamingNotificationEventType here as well?

bigludo7: I will prefer not. If we list all event type in all API that will be complicated to maintain.
Perhaps we should provide 'somewhere' a listing of all event type managed within our CAMARA APIs?

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,620 @@
...
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  #  - {}
+  - oAuth2ClientCredentials: []
+  #  - BasicAuth: []
+  #  - apiKey: []
+  - threeLegged:
+      - sim-swap:subscriptions:create
```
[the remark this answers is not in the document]
bigludo7: Agree - fixed.

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,620 @@
...
+          type: string
+          description: Detailed error description
+        
+    CreateSubscription:
+      description: The request for creating a sim swap event subscription
+      type: object
+      required:
+        - webhook
+      properties:
+        subscriptionDetail:
```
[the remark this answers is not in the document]
bigludo7: Yes got your point. As the subscription/notification is still 'fresh' I guess we're still testing it. I changed the cardinality to mandatory.

code/API_definitions/sim-swap-notification-subscription.yaml
bigludo7: Same previous

code/API_definitions/sim-swap-notification-subscription.yaml
gregory1g: But currently we list only exact one of them. API client must know what events it can receive and implement corresponding logic. One can keep the list open, but at least base type of events must be described. Events like "even notification subscription ends" and "MSISDN subscription ends" are critically important and must be included in the "base" list of event types.
All above if IMO :)

bigludo7: @gregory1g Thanks for the feedback
We have only one `type` for subscribtion but 2 `type` are managed by notification.  Indeed `org.camaraproject.sim-swap.v0.subscription-ends` is also described in this yaml. We do not plan to provide subscription for this one and send them automatically once the subscription terminates (for any reason). This is described in the doc part.

We should add in this documentation than the same resource for subscription as notification are managed for all other CAMARA APIs.

code/API_definitions/sim-swap-notification-subscription.yaml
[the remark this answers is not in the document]
fernandopradocabrillo: @bigludo7 
if we already have the security property in every operation, the global security is redundant right? I think we can remove it

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,629 @@
...
+        503:
+          $ref: "#/components/responses/Generic503"
+    delete:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Delete a sim swap event subscription'
+      operationId: deleteSubscription
+      description: delete a  given event subscription.
+      security:
+        - openId::
```
fernandopradocabrillo, line 251:
```suggestion
        - openId:

```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,629 @@
...
+paths:
+  /subscriptions:
+    post:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Create a sim swap event subscription for a phone number'
+      description: Create a sim swap event subscription for a phone number
+      operationId: createSimSwapSubscription
+      security:
+        - openId::
```
fernandopradocabrillo, line 87:
```suggestion
        - openId:

```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,629 @@
...
+        503:
+          $ref: '#/components/responses/Generic503'
+    get:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Retrieve a list of sim swap event subscription'
+      description: Retrieve a list of sim swap event subscription(s)
+      operationId: retrieveSubscriptionList
+      security:
+        - openId::
```
fernandopradocabrillo, line 166:
```suggestion
        - openId:

```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,629 @@
...
+          $ref: '#/components/responses/Generic503'
+  /subscriptions/{subscriptionId}:
+    get:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Retrieve a sim swap event subscription for a phone number'
+      operationId: retrieveSubscription
+      description: retrieve event subscription information for a given subscription.
+      security:
+        - openId::
```
fernandopradocabrillo, line 195:
```suggestion
        - openId:

```

code/API_definitions/sim-swap-notification-subscription.yaml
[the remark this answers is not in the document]
bigludo7: @fernandopradocabrillo Which one do you think we can remove ? I followed https://swagger.io/docs/specification/authentication/openid-connect-discovery/ pattern.

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,633 @@
...
+
+    It is mandatory in the subscription to provide the event `type` subscribed are several are managed in this API.
+
+    Only one event ``type`` is managed for this API: 
+      - ``org.camaraproject.sim-swap.v0.swapped`` - Event triggered when a sim swap occurs on the associated phoneNumber
+
+
+    Note: Additionnally to these list, ``org.camaraproject.sim-swa.v0.subscription-ends`` notification `type` is sent when the subscription ends. 
+    This notification did not requires dedicated subscription. 
+    It is used when the subscription expire time (required by the requester) has been reached or if the API server has to stop sending notification prematurely.
```
gregory1g, line 32: "if the API server has to stop sending notification prematurely." sounds a bit unclear, what about using more simple wording like "if the API server has to terminate subscription (e.g. when the corresponding phone number is not served by the operator anymore)"?

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,633 @@
...
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  #  - {}
+  - oAuth2ClientCredentials: []
+  #  - BasicAuth: []
+  #  - apiKey: []
+  - openId:
+      - sim-swap:subscriptions:create
```
gregory1g, line 62: Do we need independent scopes here?
From the end user privacy point of view, allow someone to subscribe to simswap is equal to allow someone to read simswap timestamp. Or, at least a permission to "subscribe" includes permission to read. Like end user revokes permission to read their SimSwap information, this logically also revokes permission to be notified about such events.  This is an API specific logic, and it is too risky to completely delegate it to Auth/Consent configuration.

bigludo7: @gregory1g This is a fair comment. Could we consider from a privacy perspective that checking now is same than checking for a coming period of x days? 
I tend to agree with you.
@fernandopradocabrillo @DT-DawidWroblewski : could you please provide your view here.

code/API_definitions/sim-swap-notification-subscription.yaml
gregory1g: ok.

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,633 @@
...
+      type : object
+      required:
+        - phoneNumber
+        - terminationReason
+      properties:
+        phoneNumber:
+          $ref: '#/components/schemas/PhoneNumber'
+        terminationReason:
+          type: string
+          enum: ["SUBSCRIPTION_EXPIRED", "NETWORK_TERMINATED"]
```
gregory1g, line 539: as mentioned above "NETWORK_TERMINTAED" covers a server side termination for whatever reason. Two questions here:
1) Is "NETWORK_TERMINTAED" a self-explaining term for a server initiated termination? One can expect that user terminated is different from network terminated.
2) shouldn't we add a possibility for an MNO to provide an optional a human readable text description explaining the reason behind server side termination?

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,633 @@
...
+
+    CloudEvent:
+      description: The notification callback
+      required:
+        - id
+        - source
+        - specversion
+        - type
+      properties:
+        id:
```
gregory1g, line 454: What about subscription ID, shouldn't it be included in the event as well?

bigludo7: Yes - It should be part of the Data structure - I add it.

code/API_definitions/sim-swap-notification-subscription.yaml
bigludo7: 1. Not sure to get it. You want me to add USER_TERMINATED ?
2. Yes make sense - I've add a `terminationDescription` attribute.

bigludo7 (commented):
Answered + updated following @gregory1g review.

code/API_definitions/sim-swap-notification-subscription.yaml
fernandopradocabrillo: and isn't it possible to have permission to read but not to subscribe to the notification?

code/API_definitions/sim-swap-notification-subscription.yaml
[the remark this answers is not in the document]
fernandopradocabrillo: @bigludo7 
We can have the security property in every operation or have it global. I think it is better to have it in every endpoint so we can specify the scope that applies to each operation properly

code/API_definitions/sim-swap-notification-subscription.yaml
gregory1g: permission to read, but not to subscribe is rather questionable, but still could be needed.
however, permission to subscribe without permission to read sounds like an dangerous security inconsistency

code/API_definitions/sim-swap-notification-subscription.yaml
gregory1g: 1. I dot think that we need to be introduce "user terminated" here. But NETWORK_TERMINTAED sounds too narrow. I suggest to consider a more generic term which covers everything which is initiated by the server side, either it is network as such, or user his/her self, or contract between application and MNO or even security conserns.

bigludo7:
Hello @gregory1g 
May I ask you to check the answer provided to your comments and let me know if some of them need additional attention. Thanks for your help.

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  #  - {}
+  - oAuth2ClientCredentials: []
+  #  - BasicAuth: []
+  #  - apiKey: []
+  - openId:
+      - sim-swap:subscriptions:create
```
fernandopradocabrillo, line 62: this hole black can be removed since we have the security property in every operation

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+  /subscriptions:
+    post:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Create a sim swap event subscription for a phone number'
+      description: Create a sim swap event subscription for a phone number
+      operationId: createSimSwapSubscription
+      security:
+        - openId:
+
```
fernandopradocabrillo, line 88:
```suggestion
```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+          $ref: '#/components/responses/Generic503'
+    get:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Retrieve a list of sim swap event subscription'
+      description: Retrieve a list of sim swap event subscription(s)
+      operationId: retrieveSubscriptionList
+      security:
+        - openId:
+
```
fernandopradocabrillo, line 168:
```suggestion
```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+  /subscriptions/{subscriptionId}:
+    get:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Retrieve a sim swap event subscription for a phone number'
+      operationId: retrieveSubscription
+      description: retrieve event subscription information for a given subscription.
+      security:
+        - openId:
+
```
fernandopradocabrillo, line 198:
```suggestion
```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+                  value:
+                    code: INVALID_INPUT
+                    status: 400
+                    message: "Schema validation failed at  ..."
+                subscriptionIdRequired:
+                  summary: eventSubscription id is required
+                  value:
+                    code: INVALID_INPUT
+                    status: 400
+                    message: "Expected property is missing: eventSubscriptionId"
```
fernandopradocabrillo, line 232: I think there is no need to include this example, I haven't seen it in other camara APIs either. But not a problem from my side if you want to keep it

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+          $ref: "#/components/responses/Generic503"
+    delete:
+      tags:
+        - Sim swap notification subscription
+      summary: 'Delete a sim swap event subscription'
+      operationId: deleteSubscription
+      description: delete a  given event subscription.
+      security:
+        - openId:
+
```
fernandopradocabrillo, line 255:
```suggestion
```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+          - sim-swap:subscriptions:delete
+      parameters:
+        - name: subscriptionId
+          in: path
+          description: Subscription identifier that was obtained from the create event subscription operation
+          required: true
+          schema:
+            type: string
```
fernandopradocabrillo, line 263: This can be moved to the components section so it can be reused in the GET too

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+                  value:
+                    code: INVALID_INPUT
+                    status: 400
+                    message: "Schema validation failed at  ..."
+                subscriptionIdRequired:
+                  summary: eventSubscription id is required
+                  value:
+                    code: INVALID_INPUT
+                    status: 400
+                    message: "Expected property is missing: eventSubscriptionId"
```
fernandopradocabrillo, line 291: same comment here

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+          tokenUrl: '{tokenUrl}'
+          scopes: {}
+    #    BasicAuth:
+    #      type: http
+    #      scheme: basic
+    #    apiKey:
+    #      type: apiKey
+    #      description: API key to authorize requests
+    #      name: apikey
+    #      in: query
```
fernandopradocabrillo, line 321: we can delete this block

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+
+    This endpoint describes the event notification received on subscription listener side when the event occurred. As for subscription, detailed description of the event notification is provided in the CAMARA API design guideline document.
+
+    **WARNING**: This callback endpoint must be exposed on the listener side as `POST /{$request.body#/webhook/notificationUrl}`
+      will be posted by the sim swap resource server to the webhook url provided by notification 
+      listener.
+
+    ## Further info and support
+
+    (FAQs will be added in a later version of the documentation)
```
fernandopradocabrillo, line 44:
```suggestion
This API provides the customer with the ability to subscribe to events related to a sim swap operation performed on an associated phone number.

    # Introduction

    This API empowers service providers and network operators with a seamless mechanism for managing and monitoring SIM card swaps, a process often used in legitimate device upgrades or replacements but also susceptible to abuse in fraudulent activities.

    # Relevant terms and definitions

    * **SIM swap**:
    A SIM swap is a process in which a mobile phone user's current SIM card is deactivated or replaced with a new one. This is typically done by contacting the user's mobile service provider and requesting a new SIM card for various reasons, such as a lost or damaged SIM card, upgrading to a new phone, or changing phone numbers while keeping the same device.

    # API Functionality

    This API allows to manage client subscriptions to notification events published by the Sim Swap resource server at the URL provided when the subscription is created.

    The API exposes following capabilities:
      * **/subscriptions**: This endpoint has two operations, one that allows to retrieve the list of Sim Swap event subscriptions and another one to create new event subscriptions.
      * **/subscriptions/{subscriptionId}**: This endpoint also has two operations, one that allows to retrieve a particular event subscription by id and another one to delete an event subscription also using the id.

    ## Sim Swap notification subscription

    Theses endpoints allow to manage notification subscription on Sim Swap event. 
    The CAMARA subscription model is detailed in the CAMARA API design guideline document and follows Cloudevents specification.

    It is mandatory in the subscription to provide the event `type` to be subscribed.

    Only one event ``type`` is managed for this API: 
      - ``org.camaraproject.sim-swap.v0.swapped`` - Event triggered when a sim swap occurs on the associated phoneNumber

    Note: Additionally to these list, ``org.camaraproject.sim-swa.v0.subscription-ends`` notification `type` is sent when the subscription ends. 
    This notification did not requires dedicated subscription. 
    It is used when the subscription expire time (required by the requester) has been reached or if the API server has to terminate subscription (e.g. when the corresponding phone number is not served by the operator anymore). 

    ### Notification callback

    This endpoint describes the event notification received on subscription listener side when the event occurred. As for subscription, detailed description of the event notification is provided in the CAMARA API design guideline document.

    **WARNING**: This callback endpoint must be exposed on the listener side as `POST /{$request.body#/webhook/notificationUrl}`
      will be posted by the sim swap resource server to the webhook url provided by notification  listener.

    ## Further info and support

    (FAQs will be added in a later version of the documentation)
```

fernandopradocabrillo: I added a few more details to the description, just a suggestion to complete a little bit more 😄

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+        phoneNumber:
+          $ref: '#/components/schemas/PhoneNumber'
+        type:
+          $ref: '#/components/schemas/SimSwapEventType'
+
+    SimSwapEventType:
+      type: string
+      description: |
+        swapped - Event triggered when a sim swap was performed for this phone number
+
```
fernandopradocabrillo, line 397:
```suggestion
```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+      description:  event structure for event subscription ends
+      allOf:
+        - $ref: '#/components/schemas/CloudEvent'
+        - type: object
+          properties:
+            data:
+              $ref: '#/components/schemas/SubscriptionEnds'
+
+    Swapped:
+      description: Event detail structure for ROAMING_ON & ROAMING_OFF event
```
fernandopradocabrillo, line 521: A copy&paste typo

```suggestion
      description: Event detail structure for SWAPPED event

```

fernandopradocabrillo (commented):
@bigludo7 
I leave you a few comments I've found in a more detailed review. Some are suggestions and some are small fixes.

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
+openapi: 3.0.3
+info:
+  title: Sim Swap Notification Subscription
```
fernandopradocabrillo, line 3:
```suggestion
  title: Sim Swap Notification Subscriptions

```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,645 @@
...
+      - sim-swap:subscriptions:delete
+servers:
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+      basePath:
+        default: /sim-swap-notification-subscriptions/v0
+        description: Base path for the sim swap notification subscription API
```
fernandopradocabrillo, line 73:
```suggestion
        description: Base path for the sim swap notification subscriptions API

```

bigludo7 (commented):
Thanks @fernandopradocabrillo for your review. I have updated the yaml accordingly (except the error message that I kept).

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -58,8 +58,10 @@ security:
   - oAuth2ClientCredentials: []
   #  - BasicAuth: []
   #  - apiKey: []
-  - threeLegged:
+  - openId:
       - sim-swap:subscriptions:create
+      - sim-swap:subscriptions:read
```
gregory1g: Using independent scopes we delegate permission consistency validation to consent management or to the end user.
When a user revokes their "read my SimSwap info" for app XYZ, it is user responsibility other ways XYZ can read their SimSwap data (e.g. via subscription).

bigludo7: Hello @gregory1g 
For me the simswap and the subscription are 2 different entities. As subscription is a distinct entity it makes sense to have distinct scope to manage it but worth to be discussed.
As we should have same discussion in subscription for device location and device status probably better to bring this discussion to commonalities. 

If a user revokes their "read my SimSwap info" for app XYZ, it is Telco commitment to terminate app XYZ subscription (if any) and prevent any sim swap data sharing with XYZ.

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,615 @@
...
+        401:
+          $ref: '#/components/responses/Generic401'
+        403:
+          $ref: '#/components/responses/Generic403'
+        409:
+          $ref: '#/components/responses/Generic409'
+        500:
+          $ref: '#/components/responses/Generic500'
+        503:
+          $ref: '#/components/responses/Generic503'
```
fernandopradocabrillo, line 146: The HTTP error codes must be in quotes. [Source](https://swagger.io/specification/v3/#responses-object)

```suggestion
        "201":
          description: Created
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/SubscriptionInfo'
        "202":
          description: Request accepted to be processed. It applies for async creation process.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/SubscriptionAsync'
        "400":
          $ref: '#/components/responses/Generic400'
        "401":
          $ref: '#/components/responses/Generic401'
        "403":
          $ref: '#/components/responses/Generic403'
        "409":
          $ref: '#/components/responses/Generic409'
        "500":
          $ref: '#/components/responses/Generic500'
        "503":
          $ref: '#/components/responses/Generic503'

```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,615 @@
...
+        400:
+          $ref: '#/components/responses/Generic400'
+        401:
+          $ref: '#/components/responses/Generic401'
+        403:
+          $ref: '#/components/responses/Generic403'
+        500:
+          $ref: '#/components/responses/Generic500'
+        503:
+          $ref: '#/components/responses/Generic503'
```
fernandopradocabrillo, line 174:
```suggestion
        "200":
          description: List of event subscription details
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/SubscriptionInfo'
        "400":
          $ref: '#/components/responses/Generic400'
        "401":
          $ref: '#/components/responses/Generic401'
        "403":
          $ref: '#/components/responses/Generic403'
        "500":
          $ref: '#/components/responses/Generic500'
        "503":
          $ref: '#/components/responses/Generic503'

```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,615 @@
...
+          schema:
+            $ref: '#/components/schemas/SubscriptionId'
+      responses:
+        200:
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SubscriptionInfo'
+        400:
```
fernandopradocabrillo, line 199:
```suggestion
        "200":
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/SubscriptionInfo'
        "400":

```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,615 @@
...
+          $ref: "#/components/responses/Generic403"
+        404:
+          $ref: "#/components/responses/Generic404"
+        500:
+          description: Server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+        503:
```
fernandopradocabrillo, line 230:
```suggestion
        "401":
          $ref: "#/components/responses/Generic401"
        "403":
          $ref: "#/components/responses/Generic403"
        "404":
          $ref: "#/components/responses/Generic404"
        "500":
          description: Server error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/ErrorInfo"
        "503":

```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,615 @@
...
+      responses:
+        204:
+          description: event subscription deleted
+        202 :
+          description: Request accepted to be processed. It applies for async deletion process.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SubscriptionAsync'
+        400:
```
fernandopradocabrillo, line 257:
```suggestion
        "204":
          description: event subscription deleted
        "202":
          description: Request accepted to be processed. It applies for async deletion process.
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/SubscriptionAsync'
        "400":
```

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,615 @@
...
+          $ref: "#/components/responses/Generic403"
+        404:
+          $ref: "#/components/responses/Generic404"
+        500:
+          description: Server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+        503:
```
fernandopradocabrillo, line 288:
```suggestion
        "401":
          $ref: "#/components/responses/Generic401"
        "403":
          $ref: "#/components/responses/Generic403"
        "404":
          $ref: "#/components/responses/Generic404"
        "500":
          description: Server error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/ErrorInfo"
        "503":
```

fernandopradocabrillo (commented):
@bigludo7 
A few format-related additional comments.

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,615 @@
...
+                - { }
+                - notificationsBearerAuth: [ ]
+
+      responses:
+        201:
+          description: Created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SubscriptionInfo'
```
fernandopradocabrillo, line 128: @bigludo7 
One doubt here, in case I've missed something, why do we have the `phoneNumber` as optional in the request body?

And, even if we have it optional in the request, shouldn't it be required in the response? I think the response must give the complete information of the subscription. Applies the same for the GET endpoints

bigludo7 (commented):
Changed cardinality for phoneNumber as suggested by @fernandopradocabrillo

code/API_definitions/sim-swap-notification-subscription.yaml
bigludo7: Yes make sense for POST & GET answers --> I've updated.

For POST request, the `phoneNumber` could be optional as we can have it by the network thru Authorization code flow or CIBA flow.

bigludo7:
Hello @fernandopradocabrillo @gregory1g @DT-DawidWroblewski Any pending issue preventing to get your approval?
We should probably moving forward with this PR and then use GitHub issue for improving the proposal. Thanks

gregory1g:
> Any pending issue preventing to get your approval?

nothing from my side.

fernandopradocabrillo:
> Any pending issue preventing to get your approval?

LGTM from my side too. If we detect something in the future we can always fix it

fernandopradocabrillo (dismissed):
LGTM

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,636 @@
...
+
+    SimSwapEventType:
+      type: string
+      description: |
+        swapped - Event triggered when a sim swap was performed for this phone number
+
+      enum:
+        - org.camaraproject.sim-swap.v0.swapped
+
+    RoamingNotificationEventType:
```
DT-DawidWroblewski, line 383: Roaming notification in sim swap API? is it a new enhancement coming with sim swap notification?

DT-DawidWroblewski (changes requested):

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,636 @@
...
+      will be posted by the sim swap resource server to the webhook url provided by notification 
+      listener.
+
+    ## Further info and support
+
+    (FAQs will be added in a later version of the documentation)
+
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: project-email@sample.com
```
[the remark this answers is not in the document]
bigludo7: Not sure - I followed here what we have in other API but this is a fair question.
My suggestion is to have an issue in Commonalities to have a common rule.

code/API_definitions/sim-swap-notification-subscription.yaml
```diff
@@ -0,0 +1,636 @@
...
+                  $ref: "#/components/responses/Generic401"
+                "403":
+                  $ref: "#/components/responses/Generic403"
+                "500":
+                  $ref: "#/components/responses/Generic500"
+                "503":
+                  $ref: "#/components/responses/Generic503"
+              security:
+                - { }
+                - notificationsBearerAuth: [ ]
```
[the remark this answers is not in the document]
bigludo7: Not sure - I've replicated same thing than in other Subscription API. If we want to provide more we should probably propose someting here: https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#security-considerations

code/API_definitions/sim-swap-notification-subscription.yaml
bigludo7: ah ah ah ... no :) good catch Dawid. I've wrongly name this class and forget to use it (2 errors on a same line). I fix it.

bigludo7 (commented):
Updated following @DT-DawidWroblewski review.

DT-DawidWroblewski (approved):
LGTM

PR #61, opened 2023-10-10
adding MoM

DT-DawidWroblewski:
#### What type of PR is this?

* documentation

adds MoMs from private repo to public


bigludo7 (approved):
Thanks Dawid
Will help you for next meeting to also capture these information

PR #64, opened 2023-10-25
MoM

DT-DawidWroblewski:
#### What type of PR is this?

* documentation

Adding MoMs


hdamker:
@DT-DawidWroblewski is there a reason why the minutes are merged into branch "MoM" and not "main"?

PR #68, opened 2023-10-31
Submission of test cases

trehman-gsma:
#### What type of PR is this?

Tests

#### What this PR does / why we need it:

Initial test case proposal. The [API readiness checklist](https://github.com/camaraproject/WorkingGroups/blob/main/Commonalities/documentation/API-Readiness-Checklist.md) states that API test cases should be created. 

Created two `.feature` files that contain Gherkin test cases. One file containing test cases for `POST /retrieve-date` and the second file containing test cases for `POST /check`.

The files were created under the `/code/Test_definitions` as per guidance in https://github.com/camaraproject/Template_Lead_Repository/pull/1

#### Which issue(s) this PR fixes:

Fixes #63 

#### Special notes for reviewers:

Initial test case proposal. Will amend as per feedback. 

#### Changelog input

```
Created /code/Test_definitions/check_simswap.feature containing test cases for POST /check

Created /code/Test_definitions/retrieve_simswap_date.feature containing test cases for POST /retrieve-date

```

PR #70, opened 2023-11-08
Telefonica test cases proposal

### LINKED ISSUE #63: Submission of test cases (Gherkin feature files)
Hi all,

As briefly discussed during the SimSwap meeting on 19/10/23 - I have a number of Gherkin test cases that I'm interested in submitting. Examples are below. Please can you take a look and advise if I should create a Pull Request? 

The Commonalities [API Readiness Checklist](https://github.com/camaraproject/WorkingGroups/blob/7baa0dd1faf4276344ba62a334f52eb06bedd751/Commonalities/documentation/API-Readiness-Checklist.md) does not yet have a reference template for test cases (it is labelled as TBD) - however I have looked through other CAMARA repositories and noted Gherkin feature files from the following projects (click to be taken to the feature files):

* [EdgeCloud (Traffic Influence)](https://github.com/camaraproject/EdgeCloud/blob/main/code/Test_definitions/Traffic_Influence_Test.feature)
* [QualityOnDemand](https://github.com/camaraproject/QualityOnDemand/blob/main/code/cucumber/src/test/resources/feature/QoD_API_Test.feature)
* [HomeDevicesQoD](https://github.com/camaraproject/HomeDevicesQoD/blob/main/code/Test_definitions/home_devices_qod_setQos.feature)

The SimSwap test cases would follow a similar pattern to the above feature files. These are some examples:
```Gherkin

Feature: Check if SIM Swap has been performed during a past period 

  @Test_Verify_Swap_True
  Scenario: Confirm that a SIM has been swapped
    Given a valid phone number "<phoneNumber>" whose SIM has been swapped in the last "<maxAge>" hours
    When a Check SimSwap request is made with a maxAge parameter of "<maxAge>"
    Then the response status code should be 200
    And the response body should confirm that a SIM swap has been performed 

    Examples:
      | phoneNumber   | maxAge |
      | +346661113334 | 24     |
      | +447748357814 | 48     |      

  @Test_Verify_Swap_True_Default_MaxAge
  Scenario: Confirm that a SIM has been swapped without specifying maxAge (should default to maxAge of 240)
    Given a valid phone number "<phoneNumber>" whose SIM has been swapped within the default max age value (240)
    When a Check SimSwap request is made without a maxAge parameter
    Then the response status code should be 200
    And the response body should confirm that a SIM swap has been performed 

  @Test_Verify_Swap_False
  Scenario: Confirm that a SIM has not been swapped
    Given a valid phone number "<phoneNumber>" whose SIM has been swapped in the last "<maxAge>" hours
    When a Check SimSwap request is made with a maxAge parameter of "<maxAge>" 
    Then the response status code should be 200
    And the response body should confirm that a SIM swap has not been performed 

    Examples:
      | phoneNumber   | maxAge |
      | +346661113334 | 24     |
      | +447748357814 | 48     |       

  @Test_Negative_MaxAge
  Scenario: Input an invalid maxAge value (less than 1)
    Given a valid phone number "<phoneNumber>" whose SIM has been swapped in the last "<maxAge>" hours
    When a Check SimSwap request is made with a maxAge parameter of "-1"
    Then the response status code should be 400
    And the response should contain an error message indicating an invalid argument
```

I have further test cases but have included the above as an example of the language and syntax. Appreciate your thoughts.

fernandopradocabrillo:
#### What type of PR is this?

* tests

Include an alternative proposal from Telefónica for the Sim swap ATP

Fixes #63 

Alternative proposal to #68 from @trehman-gsma 
Specifies a similar format but different enough to be placed in a new PR so that it can be easily reviewed.

#### Changelog input

```
Created /code/Test_definitions/check_simswap.feature containing test cases for POST /check

Created /code/Test_definitions/retrieve_simswap_date.feature containing test cases for POST /retrieve-date

```


bigludo7:
Thanks @fernandopradocabrillo 
Adding https://github.com/patrice-conil in the loop

What about one additional test for the check when the msisdn obtained from the network in the authorization first step (and passed in the token) is distinct from the one we get in the POST payload  ?

fernandopradocabrillo:
> Thanks @fernandopradocabrillo Adding https://github.com/patrice-conil in the loop
> 
> What about one additional test for the check when the msisdn obtained from the network in the authorization first step (and passed in the token) is distinct from the one we get in the POST payload ?

@bigludo7 
Regarding that, I think we should add the error explicitly in the API, as we do in other APIs, and then add the test case as well, I agree.

```
Phone number cannot be deducted from access token context.(`{"code": "SIM_SWAP.INVALID_TOKEN_CONTEXT","message": "Phone number cannot be deducted from access token context"}`)
```

fernandopradocabrillo:
@bigludo7 
Approved the PR, seems ok from my side.

For the test cases I'm considering the following:
1. The phoneNumber in the request body doesn't match the one in the access_token
2. The phoneNumber is not provided in the request body and cannot be deducted from the access token
3. The phoneNumber is provided in the request body but cannot be deducted from the access token

sounds good?

bigludo7:
> @bigludo7 Approved the PR, seems ok from my side.
> 
> For the test cases I'm considering the following:
> 
> 1. The phoneNumber in the request body doesn't match the one in the access_token
> 2. The phoneNumber is not provided in the request body and cannot be deducted from the access token
> 3. The phoneNumber is provided in the request body but cannot be deducted from the access token
> 
> sounds good?

Yes very good. Thanks

fernandopradocabrillo:
Included scenarios for HTTP 403 invalid token context

bigludo7 (approved):
LGFM
Thanks @fernandopradocabrillo

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,253 @@
...
+        And the API returns a human readable error message
+
+
+    @checkSimSwap_E19.104_PhoneNumberProvidedDoesNotMatchTheAccessToken
+    Scenario: Check that the response shows an error when phone number provided does not match the one in the access token
+        Given I want to test "checkSimSwap"
+        And a valid access_token
+        And the access token identifies the user
+        And the access token contains a phone number from the user
+        And the variable "[CONTEXT: phoneNumber]" is set to a phone number that does not belong to the Operator
```
bigludo7, line 111: Should be 
`And the variable "[CONTEXT: phoneNumber]" is set to a phone number that did not match the one encoded in the access token`
right?

fernandopradocabrillo: @bigludo7 
Yes, you are right, a copy&paste typo. I'll fix it.

Thanks!

bigludo7: @fernandopradocabrillo - I checked with @patriceConil - the grammar with use of  [ and ] and CONTEXT did not seems to be defined by Gherkin standard.

Any thoughts on this?

fernandopradocabrillo: @bigludo7 
It's a nomenclature we use in Telefónica internally. 
The concept of Context is something that exists in implementations to pass information between steps but it is not something standard in Gherkin.
We chose to indicate it this way because when creating the test, the function that implements the step can find the value "[CONTEXT: variable"]" in context.variable.

bigludo7: Thanks @fernandopradocabrillo 
This is probably an issue if we use specific nomenclature/not standard Gherkin isn't ?

Probably worth to get an agreement at Commonalities level. @rartych has logged an issue there (https://github.com/camaraproject/Commonalities/issues/94) on this topic and good to refer your work there.

trehman-gsma:
Hi @fernandopradocabrillo, 

Sorry I didn't see this earlier (notifications landed in an unmonitored email folder - my bad). Feedback below:

1 - For checkSimSwap, should we include a test case where phone number is missing? The retrieveSimSwapDate feature file has the test case already (@retrieveSimSwapDate_E19.102_NoPhoneNumber) so it aligns.

Sample:
```
    @checkSimSwap_EN.NNN_NoPhoneNumber
    Scenario: Check that the response shows an error when the phone number is missing
        Given I want to test "checkSimSwap"
        And a valid access_token
        And the request body is set to:
            """
            {}
            """
        When I request "checkSimSwap"
        Then the response status code is "400"
        And the API returns the error code "INVALID_ARGUMENT"
        And the API returns a human readable error message
```

2 - For checkSimSwap, 2 further test cases to confirm adherence to the maxAge range (min 1, max 2400). Samples below:

```
    @checkSimSwap_EN.NNN_BelowMaxAgeRange
    Scenario: Check that the response shows an error when the requested maxAge is below the minimum integer allowed
        Given I want to test "checkSimSwap"
        And a valid access_token
        And the variable "[CONTEXT: phoneNumber]" is set to an valid phone number
        And the variable "maxAge" is set below the minimum integer allowed
        And the request body is set to:
            """
            {
              "phoneNumber": "[CONTEXT: phoneNumber]"
              "maxAge": -1
            }
            """
        When I request "checkSimSwap"
        Then the response status code is "400"
        And the API returns the error code "INVALID_ARGUMENT"
        And the API returns a human readable error message

    @checkSimSwap_EN.NNN_AboveMaxAgeRange
    Scenario: Check that the response shows an error when the requested maxAge is above the maximum integer allowed
        Given I want to test "checkSimSwap"
        And a valid access_token
        And the variable "[CONTEXT: phoneNumber]" is set to an valid phone number
        And the variable "maxAge" is set above the minimum integer allowed
        And the request body is set to:
            """
            {
              "phoneNumber": "[CONTEXT: phoneNumber]"
              "maxAge": 2401
            }
            """
        When I request "checkSimSwap"
        Then the response status code is "400"
        And the API returns the error code "INVALID_ARGUMENT"
        And the API returns a human readable error message
```

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,195 @@
...
+              "phoneNumber": "[CONTEXT: phoneNumber]"
+            }
+            """
+        When I request "retrieveSimSwapDate"
+        Then the response status code is "200"
+        And the response property "$.latestSimChange" is present
+
+
+    @retrieveSimSwapDate_30.102_NoSimSwap
+    Scenario: Check SIM swap date for a valid SIM swap
```
trehman-gsma, line 182: Is this scenario wording a copy and paste error? Should it be something similar to "Check response when SIM does not have a swap date"?

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,195 @@
...
+        And the variable "[CONTEXT: phoneNumber]" is set to a phone number never connected to the network
+        And the request body is set to:
+            """
+            {
+              "phoneNumber": "[CONTEXT: phoneNumber]"
+            }
+            """
+        When I request "retrieveSimSwapDate"
+        Then the response status code is "200"
+        And the response property "$.latestSimChange" is not present
```
trehman-gsma, line 194: Has this been agreed (latestSimChange property will not be present) or is it a placeholder until there is agreement?

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,253 @@
...
+    @checkSimSwap_30.102_ValidSimSwapMaxAge
+    Scenario Outline: Check that the response shows that the SIM has been swapped
+        Given I want to test "checkSimSwap" for a user which have a phone number connected to the Operator's network and swapped in the last "<hours>" hours
+        And a valid access_token
+        And the variable "[CONTEXT: phoneNumber]" is set to a phone number connected to the network that has been swapped in the last "<hours>" hours, where "<hours>" is equal or less than provided "maxAge" request body parameter
+        And the request body is set to:
+            """
+            {
+              "phoneNumber": "[CONTEXT: phoneNumber]",
+              "maxAge": 260
```
trehman-gsma, line 190: Should it be `"maxAge": <hours>` instead of `"maxAge": 260`?

trehman-gsma:
Another suggestion. To confirm adherence to the PhoneNumber pattern, split "30.101_ValidSimSwap" into 2 test cases:
* Check SIM swap date for a valid SIM swap using a phoneNumber **with a "+" prefix**
* Check SIM swap date for a valid SIM swap using a phoneNumber **without a "+" prefix**

(This is based on experience reviewing multiple Operator deployments)

Sample:
```
@retrieveSimSwapDate_30.101_ValidSimSwapWithPlusPrefix
    Scenario: Check SIM swap date for a valid SIM swap where phone number contains a "+" prefix
        Given I want to test "retrieveSimSwapDate" for a user which have a phone number connected to the Operator's network
        And a valid access_token
        And the variable "[CONTEXT: phoneNumber]" is set to a phone number connected to the network
        And the variable "[CONTEXT: phoneNumber]" is prefixed with a "+" symbol
        And the request body is set to:
            """
            {
              "phoneNumber": "+[CONTEXT: phoneNumber]"
            }
            """
        When I request "retrieveSimSwapDate"
        Then the response status code is "200"
        And the response property "$.latestSimChange" is present

@retrieveSimSwapDate_30.102_ValidSimSwapWithoutPlusPrefix
    Scenario: Check SIM swap date for a valid SIM swap where phone number does not contain a "+" prefix
        Given I want to test "retrieveSimSwapDate" for a user which have a phone number connected to the Operator's network
        And a valid access_token
        And the variable "[CONTEXT: phoneNumber]" is set to a phone number connected to the network
        And the variable "[CONTEXT: phoneNumber]" is NOT prefixed with a "+" symbol
        And the request body is set to:
            """
            {
              "phoneNumber": "[CONTEXT: phoneNumber]"
            }
            """
        When I request "retrieveSimSwapDate"
        Then the response status code is "200"
        And the response property "$.latestSimChange" is present
```

fernandopradocabrillo:
Hi @bigludo7 @trehman-gsma !

I have updated the test cases following Commonalities guidelines.
Take a look to check if it covers everything

bigludo7 (commented):
One comment about adding 401 AUTHENTICATION_REQUIRED - this is related to the yaml but has impact on the Test Definition.
I used Location Verification TD as template.

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,176 @@
...
+    And the response property "$.message" contains a user friendly text
+
+  @check_sim_swap_401.2_expired_access_token
+  Scenario: Expired access token
+    Given the header "Authorization" is set to an expired access token
+    And the request body is set to a valid request body
+    When the request "checkSimSwap" is sent
+    Then the response status code is "401"
+    And the response property "$.status" is 401
+    And the response property "$.code" is "UNAUTHENTICATED"
```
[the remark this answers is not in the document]
fernandopradocabrillo: I'm having some trouble differenciating both scenarios to be honest haha

I don't know wich one is better, but as you said, if we go with authentication required we need to update the yaml.

Do you know exactly when each one of them applies?

image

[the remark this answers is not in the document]
bigludo7: Oh you're right there is some confusion.

I see that the our Test Definition guru (@jlurien) ;) changed to UNAUTHENTICATED (and not AUTHENTICATION_REQUIRED)[ here](https://github.com/camaraproject/DeviceLocation/pull/189/commits/ae00b2009069c3f1aa7ca15d627c57061025ca03) so please do not consider my comment, let follow what José did, and let move forward on this.

bigludo7 (approved):
LGTM

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,176 @@
...
+  @check_sim_swap_6_no_sim_swap
+  Scenario: Check that the response shows that the SIM has not been swapped
+    Given the request body property "$.phoneNumber" is set to a phone number with a SIM that has never been swapped
+    When the request "checkSimSwap" is sent
+    Then the response status code is 200
+    And the value of response property "$.swapped" == false
+
+  # Specific errors
+
+  @check_sim_swap_7_unknown_phone_number
```
trehman-gsma, line 76: Hi @fernandopradocabrillo @bigludo7,

Regarding the test cases where there is a mismatch between access token and phoneNumber - are there wider CAMARA guidelines for the expected error responses (e.g. UNKNOWN_PHONE_NUMBER / NOT_FOUND / INVALID_TOKEN_CONTEXT)? Is it based on [Commonalities' "Standardized use of CAMARA error responses"](https://github.com/camaraproject/Commonalities/blob/e25c87dc9a91054621e8ac4f1c0f4ddeecb4807b/documentation/API-design-guidelines.md#61-standardized-use-of-camara-error-responses)?

Some of the operator implementations I've experienced return a 403 in every case where there is a mismatch between access token (they do not implement the concept of a user owning multiple phone numbers at the CAMARA level), so I'm thinking these different mismatch error responses might be a little difficult to adopt across operators. If there is some guidelines I could point operators towards for these cases then that would be helpful.

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,176 @@
...
+  Scenario: Error when the phone number provided in the request body belongs to the user but does not match the phone number asssociated with the access token
+    Given the request body property "$.phoneNumber" is set to a valid testing phoneNumber that belongs to the user but does not match the one associated with the token
+    And the header "Authorization" is set to a valid access token
+    When the request "checkSimSwap" is sent
+    Then the response status code is 403
+    And the response property "$.status" is 403
+    And the response property "$.code" is "INVALID_TOKEN_CONTEXT"
+    And the response property "$.message" contains a user friendly text
+
+  @check_sim_swap_10_phone_number_provided_cannot_be_deducted_from_access_token
```
trehman-gsma, line 105: Hi @fernandopradocabrillo,

Great work on the test cases. 

Can you help me understand how a valid access token would be generated from which the phone number cannot be deduced?

In a three-legged flow, the access token is always associated to a valid phone number?

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,176 @@
...
+  @check_sim_swap_11_phone_number_not_provided_and_cannot_be_deducted_from_access_token
+  Scenario: Error when phone number can not be deducted from access token and it is not provided in the request body
+    Given the header "Authorization" is set to a valid access token from which the phone number cannot be deduced
+    When the request "checkSimSwap" is sent
+    Then the response status code is 403
+    And the response property "$.status" is 403
+    And the response property "$.code" is "INVALID_TOKEN_CONTEXT"
+    And the response property "$.message" contains a user friendly text
+
+  @check_sim_swap_12_phone_number_conflict
```
trehman-gsma, line 124: Hi @fernandopradocabrillo,

Apologies, I'm not able to comprehend this test case. Does it mean sending 2 requests using the same access token? If so, how does this result in 409?

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,176 @@
...
+  # Generic 400 errors
+
+  @check_sim_swap_400.1_invalid_phone_number
+  Scenario: Check that the response shows an error when the phone number is invalid
+    Given the request body property "$.phoneNumber" does not comply with the OAS schema at "/components/schemas/PhoneNumber"
+    When the request "checkSimSwap" is sent
+    Then the response status code is 400
+    And the response property "$.status" is 400
+    And the response property "$.code" is "INVALID_ARGUMENT"
+    And the response property "$.message" contains a user friendly text
```
trehman-gsma, line 176: What do you think about an additional test case for invalid max age (similar to @check_sim_swap_400.1_invalid_phone_number)? The purpose would be to ensure that the API complies with the minimum (1) and maximum (2400) range.

For example:

```
@check_sim_swap_400.1_invalid_max_age
  Scenario: Check that the response shows an error when the max age is invalid
    Given the request body property "$.maxAge" does not comply with the OAS schema at "/components/schemas/CreateCheckSimSwap"
    Then the response status code is 400
    And the response property "$.status" is 400
    And the response property "$.code" is "INVALID_ARGUMENT"
    And the response property "$.message" contains a user friendly text
```

trehman-gsma:
> Hi @bigludo7 @trehman-gsma !
> 
> I have updated the test cases following Commonalities guidelines. Take a look to check if it covers everything

Hi @fernandopradocabrillo - great job with the test cases. I have added a few comments to the changed files to request clarification. Thanks.

code/Test_definitions/checkSimSwap.feature
fernandopradocabrillo: Hi @trehman-gsma!

I included this test just to cover all the errors defined in the API but, actually, I don't see any use for it.
I have an open issue to discuss whether we should remove it or not #119

code/Test_definitions/checkSimSwap.feature
fernandopradocabrillo: Hi @trehman-gsma! 
Yes, the access token should always be associated to a valid phone number, but, the composition of the access token is something that is out of the scope of camara if I'm not wrong.
Maybe there is a case where the operator only includes, for example, the `sub`, so when they decode the access token they have to retrieve the phone number associated to that sub and an error may occur in that process.

To sum up, I think the way each operator handles how they validate the information associated with the token can be different and this test jsut covers a generic error case.

code/Test_definitions/checkSimSwap.feature
fernandopradocabrillo: I don't think there is a clear guideline on this I'm afraid, but I'll check with my collegues if they have more info.

I have included our view on the scenarios and this is what we are implementing in TEF. We have a more complex filtering structure in the apigw so we are able to differenciate between these cases. 

Maybe we can place a comment and clarify that depending on the auth implementations, the operator can return 404 or 403.

Edit: as stated in the issue #119 we are going to remove the unkown_phon_ number error

code/Test_definitions/checkSimSwap.feature
gregory1g: @trehman-gsma , agree. If an client requires maxAge bigger than an MNO support  it is a user error and should be considered accordingly - 4xxx.

code/Test_definitions/checkSimSwap.feature
bigludo7: @trehman-gsma May I consider this comment as resolved?

trehman-gsma: @bigludo7 I'm okay with this.

I think some operators will not be able to implement the test case - but I couldn't find any CAMARA guidelines that state whether it's mandatory that all test cases (i.e. unusual/corner cases) should be implemented.

code/Test_definitions/checkSimSwap.feature
fernandopradocabrillo: Included in the new commit

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,174 @@
...
+
+  @check_sim_swap_400.2_invalid_max_age
```
gregory1g, line 167: "does not comply with the OAS schema" covers too big and too small values, right?

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,152 @@
...
+    Given the resource "sim-swap/v0/retrieve-date"
+    And the header "Content-Type" is set to "application/json"
+    And the header "Authorization" is set to a valid access token
+    And the header "x-correlator" is set to a UUID value
+    And the request body is set by default to a request body compliant with the schema
+
+  # This first scenario serves as a minimum, not testing any specific verificationResult
+  @retrieve_sim_swap_date_1_generic_success_scenario
+  Scenario: Common validations for any sucess scenario
+    Given the request body property "$.phoneNumber" is set to a valid testing phone number
```
gregory1g, line 21: same as for the /check - all success cases for phoneNumber in the payload.

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,174 @@
...
+    And the header "Content-Type" is set to "application/json"
+    And the header "Authorization" is set to a valid access token
+    And the header "x-correlator" is set to a UUID value
+    And the request body is set by default to a request body compliant with the schema
+
+
+  # This first scenario serves as a minimum, not testing any specific verificationResult
+  @check_sim_swap_1_generic_success_scenario
+  Scenario: Common validations for any sucess scenario
+    Given the request body property "$.phoneNumber" is set to a valid testing phoneNumber
```
gregory1g, line 22: success test cases (all of them) only cover scenario where $.phoneNumber is provided. However, the recommended way to use the API is without phoneNumber in the payload. And, it looks like there is no single test case which covers this.

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,152 @@
...
+    Then the response status code is 200
+    And the response header "Content-Type" is "application/json"
+    And the response header "x-correlator" has same value as the request header "x-correlator"
+    And the response body complies with the OAS schema at "/components/schemas/SimSwapInfo"
+
+
+  # Scenarios testing specific situations
+
+  @retrieve_sim_swap_date_2_valid_sim_swap
+  Scenario: Check SIM swap date for a valid SIM swap
```
gregory1g, line 32: Should we probably use "retrieve" instead of "check" here (and for the following test cases)?

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,152 @@
...
+    Given the request body property "$.phoneNumber" is set to a phone number connected to the Operator's network
+    When the request "retrieveSimSwapDate" is sent
+    Then the response status code is 200
+    And the response property "$.latestSimChange" contains a valid timestamp
+
+  @retrieve_sim_swap_date_3_no_sim_swap_returns_activation_date
```
gregory1g, line 38: suggestion: add a comment that this test is applicable for operators which do not limit "monitoring history"

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,152 @@
...
+  Scenario: Check SIM swap date for a non-activated sim
+    Given the request body property "$.phoneNumber" is set to a phone number never connected to the Operator's network
+    When the request "retrieveSimSwapDate" is sent
+    Then the response status code is 200
+    And the response property "$.latestSimChange" is null
+
+  # This scenario applies when there is a local regulation with a time limitation on the information that can be returned
+  @retrieve_sim_swap_date_5_no_sim_swap_or_activation_date_due_to_legal_constrain
+  Scenario: Check SIM swap date for a valid SIM swap
+    Given the request body property "$.phoneNumber" is set to a valid testing phone number
```
gregory1g, line 56: Given the request body property "$.phoneNumber" is set to a valid testing phone number
=>
Given the request body property "$.phoneNumber" is set to a valid testing phone number **and SimSwap event happened before the limited history window threshold.**

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,152 @@
...
+  @retrieve_sim_swap_date_6_phone_number_not_supported
+  Scenario: Error when the service is not supported for the provided phone number
+    Given the request body property "$.phoneNumber" is set to a phone number for which the service is not available
+    When the request "retrieveSimSwapDate" is sent
+    Then the response status code is 422
+    And the response property "$.status" is 422
+    And the response property "$.code" is "NOT_SUPPORTED"
+    And the response property "$.message" contains a user friendly text
+
+  @retrieve_sim_swap_date_7_phone_number_provided_does_not_belong_to_the_user
```
gregory1g, line 72: same concern as for the corresponding /check tests

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,174 @@
...
+  Scenario: Error when the service is not supported for the provided phone number
+    Given the request body property "$.phoneNumber" is set to a phone number for which the service is not available
+    When the request "checkSimSwap" is sent
+    Then the response status code is 422
+    And the response property "$.status" is 422
+    And the response property "$.code" is "NOT_SUPPORTED"
+    And the response property "$.message" contains a user friendly text
+
+  @check_sim_swap_8_phone_number_provided_does_not_belong_to_the_user
+  Scenario: Error when provided phone number does not belong to the user
```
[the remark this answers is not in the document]
fernandopradocabrillo: For us this is a step we have prior to any other validation.
To be a valid token, the phoneNumber needs to be related to a subscriber, so we have a validation to check that indeed the phoneNumber belongs to the subscriber.
I can remove this test to not block the release, but maybe we'll like to revisit the possibility to include it in the future.

code/Test_definitions/checkSimSwap.feature
fernandopradocabrillo: yes, since we have a min and max values defined

code/Test_definitions/checkSimSwap.feature
fernandopradocabrillo: yes, I would keep the specification about the phone number tho. For example: 

```
And the header "Authorization" is set to a valid access token from which a valid testing phoneNumber can be deducted
```

wdyt? @gregory1g

gregory1g: This still keeps the main usage scenario outside of the tests coverage. With this change the only valid shown success invocation includes both access token and phoneNumber, meanwhile the recommended way to do not send phoneNumber. Probably we should focus successful tests on access token identifcation only and only keep error tests when phoneNumber does not match access token?

code/Test_definitions/checkSimSwap.feature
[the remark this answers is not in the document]
gregory1g: yes, let's remove this for now.

to be honest, I do not fully understand how this could work. To check that phoneNumber belongs to the user, you need to get the user (user id) from somewhere.  But user ID is not part of the input. Probably, you can find the user by access token, but access token is already for the exact phone number anyway. So, if the access token is generated correctly user id amd phone number "in" it must match...

code/Test_definitions/checkSimSwap.feature
gregory1g: Fully agree. phoneNumber is still needed for the service logic. But for the main success cases, it is better to move a source of this information from the request body ("$.phoneNumber" reference request body) to access token.
No sure what is the best way to express this. Probably like:

```
@check_sim_swap_1_generic_success_scenario
  Scenario: Common validations for any sucess scenario
    Given the request header "Authorization" is set to a valid access token from which a valid testing phoneNumber can be deducted
. . .

```

fernandopradocabrillo: Yep, that sounds like what I had in mind, I'll update it

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,152 @@
...
+
+  # Specific errors
+
+  @retrieve_sim_swap_date_6_phone_number_not_supported
+  Scenario: Error when the service is not supported for the provided phone number
+    Given the request body property "$.phoneNumber" is set to a phone number for which the service is not available
+    When the request "retrieveSimSwapDate" is sent
+    Then the response status code is 422
+    And the response property "$.status" is 422
+    And the response property "$.code" is "NOT_SUPPORTED"
```
sfnuser, line 69: I wonder how we can standardise the usage of `code` values across implementations? Here we have chosen `NOT_SUPPORTED` but the `spec` does not mandate this (it provides only generic error definitions with no specific values - only `examples`). I can't remember why we chose this approach. This applies to other CAMARA APIs too.

Also [Commonalities](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#61-standardized-use-of-camara-error-responses) seems to recommend the below error `code`
image

fernandopradocabrillo: The use of examples is something that comes from the beginning of CAMARA, it is the same in all APIs, I also agree it should be included as generic definition and not only examples.

Regarding the error, we haven't use that one since it only applies to APIs that use the `device` object. There is an agreement in Commonalities to use the NOT_SUPPORTED error. It is stated for example in this commonalities [PR](https://github.com/camaraproject/Commonalities/pull/145#issuecomment-2127428748). Probably also in the meeting minutes.

fernandopradocabrillo:
@gregory1g I have updated the scenarios including your comments and the new UNIDENTIFIABLE_PHONE_NUMBER error. Thanks!

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,152 @@
+Feature: CAMARA SIM Swap API, 1.0.0 - Operation retrieveSimSwapDate
+
+  # Input to be provided by the implementation to the tester
+  #
+  # Testing assets:
+  #
+  # References to OAS spec schemas refer to schemas specifies in sim_swap.yaml, version 1.0.0
+
+  Get timestamp of last MSISDN <-> IMSI pairing change for a mobile user account provided with MSIDN.
```
gregory1g, line 9: Looks a bit misleading: we do not user user accounts in camara
Suggestion:
Get timestamp of last MSISDN <-> IMSI pairing change for the provided  phone number.

code/Test_definitions/retrieveSimSwapDate.feature
```diff
@@ -0,0 +1,152 @@
...
+  @retrieve_sim_swap_date_9_phone_number_not_provided_and_cannot_be_deducted_from_access_token
+  Scenario: Error when phone number can not be deducted from access token and it is not provided in the request body
+    Given the header "Authorization" is set to a valid access token from which the phone number cannot be deducted
+    When the request "retrieveSimSwapDate" is sent
+    Then the response status code is 403
+    And the response property "$.status" is 403
+    And the response property "$.code" is "INVALID_TOKEN_CONTEXT"
+    And the response property "$.message" contains a user friendly text
+
+  @retrieve_sim_swap_date_10_phone_number_not_provided_and_cannot_be_deducted_from_access_token
```
gregory1g, line 102: conditions looks identical to @retrieve_sim_swap_date_9_phone_number_not_provided_and_cannot_be_deducted_from_access_token, but the response is different.

gregory1g (commented):
Most of the my comments are non-blocking nice to have improvement suggestions.
The one potentially blocking one is about "@retrieve_sim_swap_date_9_phone_number_not_provided_and_cannot_be_deducted_from_access_token" and "@retrieve_sim_swap_date_10_phone_number_not_provided_and_cannot_be_deducted_from_access_token"

code/Test_definitions/retrieveSimSwapDate.feature
sfnuser: @fernandopradocabrillo Thanks for the references 👍🏼

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,164 @@
...
+    And the value of response property "$.swapped" == false
+
+  # Specific errors
+
+  @check_sim_swap_7_phone_number_not_supported
+  Scenario: Error when the service is not supported for the provided phone number
+    Given the request body property "$.phoneNumber" is set to a phone number for which the service is not available
+    When the request "checkSimSwap" is sent
+    Then the response status code is 422
+    And the response property "$.status" is 422
```
sfnuser, line 81: Aren't these error returns depends on the deployment architecture? For e.g. when the service is not supported for the phone number, my ID server may not even issue an access token and can return either 401 or 403. Should we have a list of possible error codes?

code/Test_definitions/retrieveSimSwapDate.feature
fernandopradocabrillo: totally, I forgot to remove the old one, fixed!

code/Test_definitions/checkSimSwap.feature
```diff
@@ -0,0 +1,164 @@
...
+  @check_sim_swap_2_valid_sim_swap_no_max_age
+  Scenario: Check that the response shows that the SIM has been swapped
+    Given the request header "Authorization" is set to a valid access token from which a phone number connected to the Operator's network that has been swapped in the last 240 hours can be deducted
+    When the request "checkSimSwap" is sent
+    Then the response status code is 200
+    And the value of response property "$.swapped" == true
+
+  @check_sim_swap_3_valid_sim_swap_max_age
+  Scenario Outline: Check that the response shows that the SIM has been swapped
+    Given the request header "Authorization" is set to a valid access token from which a phone number connected to the Operator's network that has been swapped in the last "<hours>" hours, where "<hours>" is equal or less than provided "maxAge" request body parameter can be deducted
```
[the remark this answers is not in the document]
fernandopradocabrillo: Agree, I have split all the success scenarios into different steps, hope it is better now

code/Test_definitions/retrieveSimSwapDate.feature
fernandopradocabrillo: Yeah, I basically used the endpoint description. I don't see any problem in changing it to the one you proposed.
We can update the description in the yaml in the future

gregory1g (approved):
lgfm

bigludo7 (approved):
LGTM

PR #97, opened 2024-03-15
chore/make-response-properties-required

### LINKED ISSUE #16: Return value for swap_date when unavailable or unknown
In our implementation we have found scenarios where there is no known date for the latest SIM swapped, typically because never was a proper swap, or for other reasons. We should define the expected behaviour in this scenario. 

Current endpoint `/retrieve-date` considers the property `latest_sim_change` as optional. In the above scenario, should we return an empty object `{}` or some error, for example 404 with certain code `SIM_SWAP.NO_CHANGE`.

@bigludo7 How do you manage this scenario in your implementation? any preference?
@DT-DawidWroblewski In the MobileConnect approach, how is this handled?

### LINKED ISSUE #96: Make response properties required
**Problem description**
Make response properties "latestSimChange" and "swapped" required since they will always be returned

fernandopradocabrillo:
#### What type of PR is this?

* correction

1. Make proeperties "latestSimChange" and "swapped" required since according to agreement they will always be present in the response.
2. Update description to "retrieve-date" behaviour to reflect agreement

Fixes #96
Fixes #16

### DIFF
--- a/code/API_definitions/sim_swap.yaml
+++ b/code/API_definitions/sim_swap.yaml
@@ -30,7 +30,7 @@ info:
 
     The API provides 2 operations:
 
-    - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number.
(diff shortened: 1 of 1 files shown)


DT-DawidWroblewski (approved):
LGTM

PR #116, opened 2024-06-21
phone number instead of MSISDN to follow communalities guidelines

### LINKED ISSUE #112: Inconsistent usage of MSISDN and phoneNumber
**Problem description**
Communalities guided to use "common public" language in the API. Phone number instead of MSISDN was an example used there.
The API payload follows this guideline uses "phoneNumber", however documentation for this payload sometimes does not. This can cause some confusions (actually already did).
Example:
image

There are 2 issues here:
1) confusion for a reader: If the API creates request for a MSISDN, why the attribute is called phoneNumber and if it is the same? 
2) documentation is a place where it is especially important to use a developer friendly language (and above this is exact opposite) 

**Expected action**
Use phone number everywhere. Only use MSISDN/IMSI in the explanation what is considered as a SimSwap only.

**Additional context**

gregory1g:
#### What type of PR is this?
correction of documentation

phone number is used instead of MSISDN to follow communalities guidelines

<!-- Automatically closes linked issue when PR is merged.

Fixes #112 

#### Changelog input
phone number is used instead of MSISDN to follow communalities guidelines

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/sim_swap.yaml
+++ b/code/API_definitions/sim_swap.yaml
@@ -24,7 +24,7 @@ info:
     **SIM swap**:
     A SIM swap is a process in which a user's mobile phone number (MSISDN) is associated with a new SIM card (IMSI). This is typically done by contacting the user's mobile service provider and requesting a new SIM card for various reasons, such as a lost or damaged SIM card or upgrading to a new phone.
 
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
LGTM/

fernandopradocabrillo:
@gregory1g @bigludo7 merging since it has the 2 approvals

PR #126, opened 2024-07-17
Remove unused errors and align with commonalities errors

### LINKED ISSUE #119: Revisit defined errors
**Problem description**

We currently have a couple errors that either don't have a real use case or are not align with the rest of the CAMARA APIs.

1. Error 409 -> what is the usecase for this error? I don't think we have one, maybe we should remove it for the v0.5.0
2. Error 404 SIM_SWAP.UNKNOWN_PHONE_NUMBER-> This is the only API where we have a specific error to cover that the phone number doesn't belong to the Operator. I propose to remove it and include this case in the generic 404 NOT_FOUND for consistency.

**Expected behavior**

1. Error 409 -> being removed from the spec
2. Error 404 SIM_SWAP.UNKNOWN_PHONE_NUMBER -> being removed from the spec and included in generic 404 NOT_FOUND

fernandopradocabrillo:
#### What type of PR is this?

* correction

1. Error 409 -> removed since it doesn¡t have any use case
2. Error 404 SIM_SWAP.UNKNOWN_PHONE_NUMBER -> remove from the spec and included in generic 404 NOT_FOUND
3. Error 403 -> align with commonalities description
4. Align the rest of the errors with commonalities model

Fixes #119

### DIFF
--- a/code/API_definitions/sim_swap.yaml
+++ b/code/API_definitions/sim_swap.yaml
@@ -102,8 +102,6 @@ paths:
           $ref: "#/components/responses/Generic403"
         "404":
           $ref: "#/components/responses/Generic404"
-        "409":
-          $ref: "#/components/responses/Generic409"
         "500":
           $ref: "#/components/responses/Generic500"
         "503":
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
LGTM

PR #130, opened 2024-07-30
Include 422 UNIDENTIFIABLE_PHONE_NUMBER and info.description

### LINKED ISSUE #117: Extracting phoneNumber from the 3-legg access token
**Problem description**
When a 3-legged auth is used, the access token is MSISDN specific. This has two important consequences:
1) an API must validate that that access token matches the MSISDN in the payload and issue an error if does not
2) there is a possibility to remove MSISDN from the payload and use one extracted from the access token

As far as I understand commonalities tends to prefer the second option (avoid duplication and use the token):
https://github.com/camaraproject/Commonalities/issues/171

SimSwap does not use device object per se, however the question if one should use MSISDN from the access token instead of payload is relevant for the SimSwap. Moreover, phoneNumber is currently a mandatory parameter, therefore switching to access token based execution will be a not-backward compatible change.
 
**Possible evolution**
phoneNumber can be declared optional. Thus, it must be provided when client credentials auth is used, and must not be provided when 3-legged token is used. Or at least we can introduce such approach in the future without breaking backward compatibility.

fernandopradocabrillo:
#### What type of PR is this?

* enhancement/feature
* documentation

1. Include 422 UNIDENTIFIABLE_PHONE_NUMBER error
2. Add description of error scenarios to info.description section

Fixes #117

### DIFF
--- a/code/API_definitions/sim_swap.yaml
+++ b/code/API_definitions/sim_swap.yaml
@@ -40,6 +40,30 @@ info:
 
     It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.
 
+    # Identifying a phone number from the access token
+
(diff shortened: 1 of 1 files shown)


hdamker (approved):
LGTM - let me know if you need support to get it merged.

fernandopradocabrillo:
> LGTM - let me know if you need support to get it merged.

@hdamker yes please, since Ludovic is OOO I'm the only codeowner left haha Thanks!

PR #146, opened 2024-08-22
Remove obsolete/wrong statement in API description that "either one or both endpoints could be implemented"

### LINKED ISSUE #144: Obsolete/wrong statement in API description that "either one or both endpoints could be implemented"
**Problem description**

The API description contains (still?) the sentence:

> Depending on the network provider implementation, legal enforcement, etc... either one or both endpoints could be implemented.

But none of the both endpoints has 501 as valid response defined, and my understanding is that the design of the responses is now done in a way that every network provider can implement both endpoints, and must do so if offering the API.

**Expected action**

Delete the sentence from the description.

**Additional context**

fernandopradocabrillo:
#### What type of PR is this?

* correction

Remove obsolete/wrong statement in API description that "either one or both endpoints could be implemented"

Fixes #144

### DIFF
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -17,8 +17,6 @@ info:
     * When did the last SIM swap occur?
     * Has a SIM swap occurred during last n hours?
 
-    Depending on the network provider implementation, legal enforcement, etc... either one or both endpoints could be implemented.
-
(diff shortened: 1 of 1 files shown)


hdamker (approved):

bigludo7 (approved):
LGTM

PR #153, opened 2024-09-18
Add management of monitoredPeriod in sim swap for retrieve-date operation

### LINKED ISSUE #124: Add MonitoredPeriod into the API Response if a Telco cannot show the date / data for for Privacy Reasons
**Problem / Context** 

- I noticed that when doing a SIM Swap API request against a number that had a SIM Swap event greater than the maxAge that the Telco wants to expose for privacy reasons, the response comes back `blank` but still logs as a `200` API successful event 
- In this case, for end users it would be better for an `error` to be returned saying something like; `Requests for data more than x days not allowed` 
- Not all telcos are using the same MaxAge time period at the moment, this will be due to privacy regulations that differ for each country

**To Reproduce**
-  Perform SIM Swap /retrieve-date and /check operations on a German number with Telco (DT) that had the events done more than 30 days ago
- We get back blank data points and 200 API response 

**Expected behavior**
As the restriction on the data being sent back is due to the Telcos privacy reasons, it seems it would be better for:
- An error API response to be returned, as a [451](https://en.wikipedia.org/wiki/HTTP_451) (unavailable for legal reasons)
- An error message to be returned instead of a blank, i.e.  `Requests for data more than x days not allowed`

bigludo7:
#### What type of PR is this?

* enhancement/feature

Add monitoredPeriod attribute management.

Fixes #124 

#### Changelog input

```
 release-note
- Add management of the `monitoredPeriod` in the retrieve-date operation response.

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -31,7 +31,7 @@ info:
     - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number.
 
       - If no swap has been performed, the API will return the SIM activation date (the timestamp of the first time that the SIM connected to the network).
-    
+
       - If the latest SIM swap date (or the activation date if no SIM swap) cannot be communicated due to local regulations preventing the safekeeping of the information for longer than the stated period, a `null` value will be returned. Optionally, a `monitoredPeriod` could be provided to indicate that no SIM swap has occurred during this timeframe (in days).
 
     - POST check: Checks if SIM swap has been performed during a past period (defined in the request with 'maxAge' attribute) for a given phone number.
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -28,7 +28,11 @@ info:
 
     The API provides 2 operations:
 
-    - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number. If no swap has been performed, the API will return the SIM activation date (the timestamp of the first time that the sim connected to the network) by default, unless this is not possible due to local regulations preventing the safekeeping of the information for longer than the stated period of time. If this is the case, a `null` value will be returned.
+    - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number.
+
+      - If no swap has been performed, the API will return the SIM activation date (the timestamp of the first time that the SIM connected to the network). 
+      
+      - If the latest SIM swap date (or the activation date if no SIM swap) cannot be communicated due to local regulations preventing the safekeeping of the information for longer than the stated period, a `null` value will be returned. Optionally, a `monitoredPeriod` could be provided to indicate that no SIM swap has occurred during this timeframe (in days).
 
     - POST check: Checks if SIM swap has been performed during a past period (defined in the request with 'maxAge' attribute) for a given phone number.
 
@@ -73,13 +77,13 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 1.0.0
+  version: wip
   x-camara-commonalities: 0.4.0
 externalDocs:
   description: Product documentation at Camara
   url: https://github.com/camaraproject/SimSwap
 servers:
-  - url: "{apiRoot}/sim-swap/v1"
+  - url: "{apiRoot}/sim-swap/wip"
     variables:
       apiRoot:
         default: http://localhost:9091
@@ -116,6 +120,13 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/SimSwapInfo"
+              examples:
+                RETRIEVE_DATE:
+                  $ref: "#/components/examples/RETRIEVE_DATE"
+                RETRIEVE_MONITORED_PERIOD:
+                  $ref: "#/components/examples/RETRIEVE_MONITORED_PERIOD"
+                RETRIEVE_MONITORED_NULL:
+                  $ref: "#/components/examples/RETRIEVE_MONITORED_NULL" 
         "400":
           $ref: "#/components/responses/Generic400"
         "401":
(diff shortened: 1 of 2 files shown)


fernandopradocabrillo (dismissed):

code/API_definitions/sim-swap.yaml
```diff
@@ -208,6 +219,10 @@ components:
           description: Timestamp of latest SIM swap performed. It must follow [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) and must have time zone. Recommended format is yyyy-MM-dd'T'HH:mm:ss.SSSZ (i.e. which allows 2023-07-03T14:27:08.312+02:00 or 2023-07-03T12:27:08.312Z)
           nullable: true
           example: "2023-07-03T14:27:08.312+02:00"
+        monitoredPeriod:
+          type: integer
+          description: Timeframe in days for SIM card change supervision for the phone number. It could be valued in the response if the latest SIM swap occurred before this monitored period.
```
gregory1g, line 224: question: could it make sense to use hours instead of days here, to align this attribute with "maxAge" parameter from the "check" operation?
When I use check operation , maxAge is internally compared to the monitoredPeriod and the API returns a error when maxAge >monitoredPeriod.
From this perspective it could make sense to use same UoM for both. 
On the other hand monitoredPeriod represented in hours promises to be a big number.
Would like to hear other opinions here.

code/API_definitions/sim-swap.yaml
```diff
@@ -28,7 +28,11 @@ info:
 
     The API provides 2 operations:
 
-    - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number. If no swap has been performed, the API will return the SIM activation date (the timestamp of the first time that the sim connected to the network) by default, unless this is not possible due to local regulations preventing the safekeeping of the information for longer than the stated period of time. If this is the case, a `null` value will be returned.
+    - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number.
+
+      - If no swap has been performed, the API will return the SIM activation date (the timestamp of the first time that the SIM connected to the network).
+
+      - If the latest SIM swap date (or the activation date if no SIM swap) cannot be communicated due to local regulations preventing the safekeeping of the information for longer than the stated period, a `null` value will be returned. Optionally, a `monitoredPeriod` could be provided to indicate that no SIM swap has occurred during this timeframe (in days).
```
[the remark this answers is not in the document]
bigludo7: Applied

code/API_definitions/sim-swap.yaml
bigludo7: I asked myself same question when i proposed this design. It seems to me that the use of day is more aligned with this attribute meaning and I preferred it that unit consistency across the API. 

BTW Thanks @gregory1g for your helpful review :)

bigludo7:
I need your help on this one @rartych !
I do not see how to solve the linting issue.
Thanks

code/API_definitions/sim-swap.yaml
```diff
@@ -28,7 +28,11 @@ info:
...
+    - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number.
+
+      - If no swap has been performed and the MNO supports unlimited SimSwap monitoring timeframe, the API will return the SIM activation date (the timestamp of the first time that the SIM connected to the network).
+
+      - If the latest SIM swap date (or the activation date if no SIM swap) cannot be communicated due to local regulations (or MNO internal privacy policies) preventing the safekeeping of the information for longer than the stated period, a `null` value will be returned. Optionally, a `monitoredPeriod` could be provided to indicate monitored time frame (in days) supported by the MNO. In this case the response must be treated as "there was no sim swap events during 'monitoredPeriod'. Either the parameter is optional, it is recommended to support it in SimSwap implementations.
```
gregory1g, line 35: "there was no sim swap events" -> "there were no sim swap events"?

code/API_definitions/sim-swap.yaml
gregory1g: Agree, even be not fully aligned with maxAge,  "days" still looks better here.

code/API_definitions/sim-swap.yaml
```diff
@@ -30,9 +30,9 @@ info:
 
     - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number.
 
-      - If no swap has been performed, the API will return the SIM activation date (the timestamp of the first time that the SIM connected to the network).
+      - If no swap has been performed and the MNO supports unlimited SimSwap monitoring timeframe, the API will return the SIM activation date (the timestamp of the first time that the SIM connected to the network).
```
hdamker, line 33: Please avoid "MNO"

```suggestion
      - If no swap has been performed and the network operator supports unlimited SimSwap monitoring timeframe, the API will return the SIM activation date (the timestamp of the first time that the SIM connected to the network).
```

code/API_definitions/sim-swap.yaml
bigludo7: Fixed

code/API_definitions/sim-swap.yaml
```diff
@@ -73,13 +77,13 @@ info:
...
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 1.0.0
+  version: wip
   x-camara-commonalities: 0.4.0
 externalDocs:
   description: Product documentation at Camara
   url: https://github.com/camaraproject/SimSwap
 servers:
-  - url: "{apiRoot}/sim-swap/v1"
+  - url: "{apiRoot}/sim-swap/wip"
```
fernandopradocabrillo, line 86: @bigludo7  shouldn't this be vwip?

fernandopradocabrillo:
@bigludo7 there seems to be an error in the linting but I don't find what it is

bigludo7:
> @bigludo7 there seems to be an error in the linting but I don't find what it is

Yes - It's because I have 'null' in some example. @rartych find the cause but he hasn't yet the resolution. If we did not find it I will remove the example but as we're not particularly on rush we can let some time to Rafal to find a trick.

code/API_definitions/sim-swap.yaml
bigludo7: @fernandopradocabrillo Yes - this is strange because I've fixed this one in https://github.com/camaraproject/SimSwap/pull/153/commits/7acbc2c4caba695cecf460fb1322dfe2bc7e5c2e . Could you check if it's good now? Thanks

fernandopradocabrillo: In that commit I also see it as wip. Let me place the suggestion so we can apply directly

```suggestion
  - url: "{apiRoot}/sim-swap/vwip"
```

bigludo7:
@fernandopradocabrillo @gregory1g @QaunainM  As we're not preparing the Spring25 release we can move forward on this one.
Welcome to get your approval to merge this one.
Thanks

fernandopradocabrillo (approved):

QaunainM:
@bigludo7 for this item, I wonder why having a `monitredPeriod; {Interger: amount of days} ` in the API response is optional, because otherwise we will still get a 200 with a blank response

PR #159, opened 2024-10-17
Update test plan

### LINKED ISSUE #154: Issue with @check_sim_swap_6_no_sim_swap scenario
**Problem description**
Doing the test I have an issue with this scenario:

```
  @check_sim_swap_6_no_sim_swap
  Scenario: Check that the response shows that the SIM has not been swapped
    Given the request header "Authorization" is set to a valid access token from which a phone number connected to the Operator's network can be deducted
    And the SIM for this phone number has never been swapped
    And the value of response property "$.swapped" == false
```

Indeed, we consider that: `New subscription is considered as a SIM swap as well` in the yaml so for the test must be reworked accordingly.
Suppose a new line was activated Oct8th at 4pm
If I make a query Oct9th at 4pm

- I expect true if `maxAge` is set to 36
- I expect false if `maxAge` is set to 12

**Expected behavior**
First get @fernandopradocabrillo feedback :)

**Alternative solution**

**Additional context**

fernandopradocabrillo:
#### What type of PR is this?

* correction
* tests

1. Includes a new test case to check the response when there is no sim swap providing maxAge and not providing it
2. Change Given step to align with other test plans how the phoneNumber is informed
3. Rename files to comply with commonalities guidelines
4. Remove 403 INVALID_TOKEN_CONTEXT test case that should not be covered by this error code

Fixes #154


bigludo7 (approved):
Look good for me

PR #168, opened 2024-10-30
Update sim-swap-check.feature

### LINKED ISSUE #167: Issue with test definition @check_sim_swap_7_no_sim_swap_with_max_age
**Problem description**
The test states that:
```  Scenario Outline: Check that the response shows that the SIM has not been swapped - maxAge is provided in the request parameter
    Given a valid phone number identified by the token or provided in the request body
    And the SIM for this phone number has never been swapped
    And the activation of the SIM occurred more than "<hours>" hours ago
    And the request body property "maxAge" is set to a value equal or greater than "<hours>" within the allowed range
    And the value of response property "$.swapped" == false
```

but it should be:

```  Scenario Outline: Check that the response shows that the SIM has not been swapped - maxAge is provided in the request parameter
    Given a valid phone number identified by the token or provided in the request body
    And the SIM for this phone number has never been swapped
    And the activation of the SIM occurred more than "<hours>" hours ago
    And the request body property "maxAge" is set to a value less than "<hours>" within the allowed range
    And the value of response property "$.swapped" == false
```

**Expected behavior**
Fix the test case

**Alternative solution**

**Additional context**

bigludo7:
Fix #167

#### What type of PR is this?

* correction

Fix test definition @check_sim_swap_7_no_sim_swap_with_max_age

Fixes #167 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/Test_definitions/sim-swap-check.feature
+++ b/code/Test_definitions/sim-swap-check.feature
@@ -82,7 +82,7 @@ Feature: CAMARA SIM Swap API, 1.0.0 - Operation checkSimSwap
     Given a valid phone number identified by the token or provided in the request body
     And the SIM for this phone number has never been swapped
     And the activation of the SIM occurred more than "<hours>" hours ago
-    And the request body property "maxAge" is set to a value equal or greater than "<hours>" within the allowed range
+    And the request body property "maxAge" is set to a value less than "<hours>" within the allowed range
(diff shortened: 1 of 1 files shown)


fernandopradocabrillo (approved):
Thanks!

gregory1g:
agree.

PR #170, opened 2024-11-05
Update test wording to avoid confusion with "activation" terminology

### LINKED ISSUE #169: Rewording test case to avoid confusion with "activation" term
**Problem description**

Update following test scenario to avoid confusion with "activation" terminology

```
@retrieve_sim_swap_date_4_sim_not_activated
  Scenario: Retrieves SIM swap date for a non-activated sim
    Given a valid phone number identified by the token or provided in the request body
    And the sim for that device has never been connected to the Operator's network
    When the request "retrieveSimSwapDate" is sent
    And the response property "$.latestSimChange" is null
```

with

```
@retrieve_sim_swap_date_4_sim_never_connected
  Scenario: Retrieves SIM swap date for a sim that has never been connected to the network
    Given a valid phone number identified by the token or provided in the request body
    And the sim for that phone number has never been connected to the Operator's network
    When the request "retrieveSimSwapDate" is sent
    And the response property "$.latestSimChange" is null
```

fernandopradocabrillo:
#### What type of PR is this?

* correction
* tests

Updates test @retrieve_sim_swap_date_4_*  wording

Fixes #169

### DIFF
--- a/code/Test_definitions/sim-swap-retrieveDate.feature
+++ b/code/Test_definitions/sim-swap-retrieveDate.feature
@@ -44,11 +44,11 @@ Feature: CAMARA SIM Swap API, 1.0.0 - Operation retrieveSimSwapDate
     Then the response status code is 200
     And the response property "$.latestSimChange" contains the sim's activation timestamp
 
-  # This test applies if the operator allows to do the request for a sim that has never been connected to the network
-  @retrieve_sim_swap_date_4_sim_never_connected
-  Scenario: Retrieves SIM swap date for a sim that has never been connected to the network
+  # This test applies if the operator allows to do the request for a phone number that hasn't been associated with a sim card yet
+  @retrieve_sim_swap_date_4_sim_never_associated
+  Scenario: Retrieves SIM swap date for a phone number that has never been associated with a sim card
     Given a valid phone number identified by the token or provided in the request body
-    And the sim for that phone number has never been connected to the Operator's network
+    And the phone number is not associated to any sim card
     When the request "retrieveSimSwapDate" is sent
     Then the response status code is 200
     And the response property "$.latestSimChange" is null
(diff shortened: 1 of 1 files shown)


code/Test_definitions/sim-swap-retrieveDate.feature
```diff
@@ -45,10 +45,10 @@ Feature: CAMARA SIM Swap API, 1.0.0 - Operation retrieveSimSwapDate
     And the response property "$.latestSimChange" contains the sim's activation timestamp
 
   # This test applies if the operator allows to do the request for a sim that has never been connected to the network
-  @retrieve_sim_swap_date_4_sim_not_activated
-  Scenario: Retrieves SIM swap date for a non-activated sim
+  @retrieve_sim_swap_date_4_sim_never_connected
+  Scenario: Retrieves SIM swap date for a sim that has never been connected to the network
```
gregory1g, line 49: The test case looks a bit fuzzy for me. Sim was not connected to the network - ok, but was it replaced? One can purchase a SIM and request an replacement before it is connect it to the network. From the SimSwap definition (and potential attack vector) this is still a SimSwap.  

Suggestion: avoid "connected to the network" validation and keep things simple: Sim either was associated with a phone number (pre-provision, on first-call, different ways) or not. The date when this association is the SimSwap date, independent on what user did with that or other SIM cards associated with the phone number.

bigludo7: I'm fine with the proposal to change wording from _connected_ to _associated_ if it works also for @fernandopradocabrillo

fernandopradocabrillo: Yes I agree with the proposal. It is clearer and makes more sense that way, I'll adapt it
Thanks!

bigludo7 (approved):
LGTM

PR #179, opened 2024-12-23
Sim swap subscriptions alignement with commonalities 0.5

### LINKED ISSUE #161: Scope for Spring25 release (in preparation)
**Problem description**
This issue should describe the scope for the next versions of the APIs within the SimSwap, targeting the Spring25 meta-release.

**Expected action**
### Target versions (tbd)
sim-swap 1.1 (_or 2.0 if new item with breaking change_)
sim-swap-subscriptions 0.2

### Issues with fixes for Spring25 scope

**sim-swap:**
- [x] Add MonitoredPeriod into the API Response if a Telco cannot show the date / data for for Privacy Reasons  #124 
- [x]  Update test wording to avoid confusion with "activation" terminology #170 
- [x] Fix test definition @check_sim_swap_7_no_sim_swap_with_max_age #168 
- [ ] Change error structure as defined in  https://github.com/camaraproject/Commonalities/pull/316 
- [ ] Enhancements and Alignments in Errors https://github.com/camaraproject/Commonalities/pull/329 & https://github.com/camaraproject/Commonalities/pull/324 
- [ ] Remove 500, 503 & 504 errors from the yaml as stated here: https://github.com/camaraproject/Commonalities/issues/321#issuecomment-2429158445  - this topic is still open [here](https://github.com/camaraproject/Commonalities/issues/355) and I suggest to wait discussion outcome before to proceed. 

**sim-swap-subscriptions:**
- [x] Add TerminationReason SUBSCRIPTION_DELETED #177 
- [x] remove allof in sinkcredential #175 

#160 in particular:
- [ ] Error alignments for Event Notifications Artifact - https://github.com/camaraproject/Commonalities/pull/351
- [ ] Align subscription model as defined in https://github.com/camaraproject/Commonalities/pull/313 

**Test definitions:**
- [ ] Align API-Testing-Guidelines with new Enhancements and Alignments in Errors [here](https://github.com/camaraproject/Commonalities/pull/343)

**Additional context**

bigludo7:
#### What type of PR is this?

* enhancement/feature

Alignement of sim_swap-subscriptions.yaml & sim_swap-subscriptions.feature with commonalities 0.5 rules.
See detail in release note.

Fixes #161 

This is breaking change.

#### Changelog input

```
 release-note
sim_swap-subscriptions.yaml:
- Change error structure definition to normalize error & status
- Update error code to introduce 429 & 422 - removed 5xx, 410 and 415 error
- Follow principle of data minimization by removing sinkCredential in POST & GET response
- Adapt request for 3-legs usage featuring identification of the phoneNumber
- Improve documentation part
- Add 3-legs examples

sim_swap-subscriptions.feature:
- Align test scenario with commonalities 0.5 improvements

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -1330,7 +1330,8 @@ components:
         types:
           - org.camaraproject.sim-swap-subscriptions.v0.swapped
         config:
-          subscriptionDetail: null
+          subscriptionDetail: {}
+
           subscriptionExpireTime: '2025-01-17T13:18:23.682Z'
         id: '1119920371'
         startsAt: '2024-06-07T16:10:15.302Z'
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -1307,7 +1307,7 @@ components:
         types:
           - org.camaraproject.sim-swap-subscriptions.v0.swapped
         config:
-          subscriptionDetail: null
+          subscriptionDetail: {}
           subscriptionExpireTime: '2025-01-17T13:18:23.682Z'
     SUBSCRIPTION_RESPONSE:
       value:
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -51,9 +51,8 @@ info:
       #### phoneNumber presence:
       These rules apply for subscription:
 
-      - If 3-legged access token is used, the POST and GET responses must not provide any device identifier (``phoneNumber``).
-      - If 2-legged access token is used, the presence of a device identifier (``phoneNumber``) in the response is mandatory and should be
-       the same identifier type than the one provided in the request.
+      - If 3-legged access token is used, the POST and GET{id} requests & responses must not provide any device identifier (``phoneNumber``). For GET (list) only subscription(s) for this exactly phone number is/are retrieved (implicit filtering).
+      - If 2-legged access token is used, for POST and GET{id}, the presence of a device identifier (``phoneNumber``) in the request & response is mandatory. For GET (list) the device identifier (``phoneNumber``) could be used as filtering criteria.
 
       #### sinkCredential presence:
 
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -1078,7 +1078,6 @@ components:
                   code:
                     enum:
                       - PERMISSION_DENIED
-                      - INVALID_TOKEN_CONTEXT
                       - SUBSCRIPTION_MISMATCH
           examples:
             GENERIC_403_PERMISSION_DENIED:
@@ -1087,12 +1086,6 @@ components:
                 status: 403
                 code: PERMISSION_DENIED
                 message: Client does not have sufficient permissions to perform this action.
-            GENERIC_403_INVALID_TOKEN_CONTEXT:
-              description: Reflect some inconsistency between information in some field of the API and the related OAuth2 Token
-              value:
-                status: 403
-                code: INVALID_TOKEN_CONTEXT
-                message: "{{field}} is not consistent with access token."
             GENERIC_403_SUBSCRIPTION_MISMATCH:
               description: Inconsistent access token for requested subscription
               value:
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -1358,7 +1358,8 @@ components:
           types:
             - org.camaraproject.sim-swap-subscriptions.v0.swapped
           config:
-            subscriptionDetail: null
+            subscriptionDetail: {}
+
             subscriptionExpireTime: '2025-01-17T13:18:23.682Z'
           id: '1119920371'
           startsAt: '2024-06-07T16:10:15.302Z'
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -51,8 +51,8 @@ info:
       #### phoneNumber presence:
       These rules apply for subscription:
 
-      - If 3-legged access token is used, the POST and GET{id} requests & responses must not provide any device identifier (``phoneNumber``). For GET (list) only subscription(s) for this exactly phone number is/are retrieved (implicit filtering).
-      - If 2-legged access token is used, for POST and GET{id}, the presence of a device identifier (``phoneNumber``) in the request & response is mandatory. For GET (list) the device identifier (``phoneNumber``) could be used as filtering criteria.
+      - If 3-legged access token is used, the POST and GET{id} requests & responses must NOT provide thev``phoneNumber``. For GET (list) only subscription(s) for this exactly phone number is/are retrieved (implicit filtering).
+      - If 2-legged access token is used, for POST and GET{id}, the presence of ``phoneNumber`` in the request & response is mandatory. For GET (list) the device identifier (``phoneNumber``) could be used as filtering criteria.
 
       #### sinkCredential presence:
 
@@ -1238,13 +1238,13 @@ components:
               value:
                 status: 429
                 code: QUOTA_EXCEEDED
-                message: Either out of resource quota or reaching rate limiting.
+                message: Rejected due to exceeding a business quota limit.
             GENERIC_429_TOO_MANY_REQUESTS:
               description: API Server request limit is overpassed
               value:
                 status: 429
                 code: TOO_MANY_REQUESTS
-                message: Either out of resource quota or reaching rate limiting.
+                message: Rejected due to request rate limit overpassed.
   examples:
     SWAPPED:
       value:
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -1307,7 +1307,7 @@ components:
         types:
           - org.camaraproject.sim-swap-subscriptions.v0.swapped
         config:
-          subscriptionDetail:
+          subscriptionDetail: null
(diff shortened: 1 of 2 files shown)


bigludo7:
Hello @rartych 
I will need your help for the megalinter issue.

Megalinter is unhappy because in the 3LEGS examples the `subscriptionDetail` is null.
This attribute is mandatory but in these examples we will have an empty object.

I tried
`subscriptionDetail: null`
or
`subscriptionDetail: `

but neither work. Any idea?

Thanks for you help.

code/API_definitions/sim-swap-subscriptions.yaml
```diff
@@ -1164,6 +1304,15 @@ components:
           accessToken: "yJ2ZXIiOiIxLjAiLCJ0eXAiO..."
           accessTokenExpiresUtc: "1717753038"
           accessTokenType: "bearer"
+        types:
+          - org.camaraproject.sim-swap-subscriptions.v0.swapped
+        config:
+          subscriptionDetail: null
```
rartych, line 1310:
```suggestion
          subscriptionDetail: {}
```

code/API_definitions/sim-swap-subscriptions.yaml
```diff
@@ -1194,3 +1350,16 @@ components:
...
           expiresAt: '2024-06-07T16:10:15.302Z'
           status: ACTIVATION_REQUESTED
+    SUBSCRIPTIONS_3LEGS:
+      value:
+        - protocol: HTTP
+          sink: https://endpoint.example.com/sink
+          types:
+            - org.camaraproject.sim-swap-subscriptions.v0.swapped
+          config:
+            subscriptionDetail: null
```
rartych, line 1360:
```suggestion
            subscriptionDetail: {}

```

code/API_definitions/sim-swap-subscriptions.yaml
```diff
@@ -1174,16 +1323,23 @@ components:
...
         status: ACTIVATION_REQUESTED
-
+    SUBSCRIPTION_RESPONSE_3LEGS:
+      value:
+        protocol: HTTP
+        sink: https://endpoint.example.com/sink
+        types:
+          - org.camaraproject.sim-swap-subscriptions.v0.swapped
+        config:
+          subscriptionDetail: null
```
rartych, line 1333:
```suggestion
          subscriptionDetail: {}

```

rartych: @bigludo7 I guees the `subscriptionDetail: {}` construction should work.

bigludo7: Thank you @rartych  !!! it works - you rock !!

code/Test_definitions/sim-swap-subscriptions.feature
```diff
@@ -63,7 +63,7 @@ Feature: CAMARA sim swap subscriptions  API, v0.1.1
   Scenario: Check existing subscription is retreived in list
     Given valid subscriptions are existing
     And use BaseURL
-    When get sim swap subscription
+    When the HTTP "GET" request is sent
```
gregory1g, line 66: This PQ makes response depending on the token type. As a result statement "And the response body complies with an array of OAS schema defined at "#/components/schemas/Subscription""  does not describe the expected result clearly enough anymore. Separate test cases are needed for 2 and 3 legged requests.

And for 3-legged case one need to specify which exactly subscriptions must be returned.

code/API_definitions/sim-swap-subscriptions.yaml
```diff
@@ -1194,3 +1351,17 @@ components:
           startsAt: '2024-06-07T16:10:15.302Z'
           expiresAt: '2024-06-07T16:10:15.302Z'
           status: ACTIVATION_REQUESTED
+    SUBSCRIPTIONS_3LEGS:
```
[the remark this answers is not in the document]
bigludo7: This is a good comment. I propose to update the documentation part:
_______________________________________________________________________
These rules apply for subscription:

If 3-legged access token is used:
- For POST and GET{id} request & responses must not provide any device identifier `phoneNumber`.
- For GET (list) only subscription for this exactly phone number is retrieved (implicit filtering)

If 2-legged access token is used:
- For POST and GET{id} the presence of a device identifier `phoneNumber` in the request t& response is mandatory. 
-  For GET (list) the `phoneNumber` could be used as filtering criteria. 

_______________________________________________________________________

code/Test_definitions/sim-swap-subscriptions.feature
bigludo7: Yes make sense
I've reworked the proposal with @sim_swap_subscription_retrieve_04_retrieve_list_2legs, @sim_swap_subscription_retrieve_07_retrieve_list_3legs and @sim_swap_subscription_retrieve_08_retrieve_empty_list_3legs

code/API_definitions/sim-swap-subscriptions.yaml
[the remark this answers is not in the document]
bigludo7: I got your point for the GET (list) but I'm wondering if we have to strictly enforce that MNO **must** support client credential. I will prefer to keep it as a recommandation and probably waiting for live implementation to get feedbacks. If we enforce it now it will be tough from a certification perspective.

bigludo7:
@fernandopradocabrillo @gregory1g @rartych  Following https://github.com/camaraproject/Commonalities/issues/369 discussion I think it's fair to remove 403 INVALID_TOKEN_CONTEXT example in the yaml.

May I ask you to review this new proposal soon as we have to prepare the RC candidate before Jan17th. We can still modify after but we have to achieve the milestone.

code/API_definitions/sim-swap-subscriptions.yaml
```diff
@@ -46,14 +46,40 @@ info:
       will be posted by the sim swap resource server to the webhook url provided by notification
       listener.
 
+    ### Data Minimization
+
+      #### phoneNumber presence:
+      These rules apply for subscription:
+
+      - If 3-legged access token is used, the POST and GET{id} requests & responses must not provide any device identifier (``phoneNumber``). For GET (list) only subscription(s) for this exactly phone number is/are retrieved (implicit filtering).
+      - If 2-legged access token is used, for POST and GET{id}, the presence of a device identifier (``phoneNumber``) in the request & response is mandatory. For GET (list) the device identifier (``phoneNumber``) could be used as filtering criteria.
```
gregory1g, line 55: This is a very short description and it keeps some questions opened.
for example:
1) it says that one can use phoneNumber filtering criteria. But the GET "/subscriptions" specification does not support filtering.
2)  what GET subscription returns if one does not use filtering criteria in 2-legged token case?
3) what happens if one uses 3-legged token for "GET /subscriptions/{subscriptionId}", where  the target subscription which is not associated with a phoneNumber from the token?
4) what happens if one tries to use 2-legged token towards a subscription created by another API invoker?

code/Test_definitions/sim-swap-subscriptions.feature
```diff
@@ -53,28 +53,51 @@ Feature: CAMARA sim swap subscriptions  API, v0.1.1
...
     Then the response property "$.status" is 200
     And the response header "Content-Type" is "application/json"
     And the response header "x-correlator" has same value as the request header "x-correlator"
     And the response body complies with an array of OAS schema defined at "#/components/schemas/Subscription"
     And all valid subscriptions are listed
 
+  @sim_swap_subscription_retrieve_07_retrieve_list_3legs
+  Scenario: Check existing subscriptions are retrieved in list
+    Given valid subscription is existing for a phoneNumber
```
gregory1g, line 75: How "valid subscription" is defined and how one can list "not valid" subscription to, for example, delete them?

code/API_definitions/sim-swap-subscriptions.yaml
bigludo7: @gregory1g Fair questions. We have added a piece in commonalities [her](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#117-resource-access-restriction)e to target your questions 2. 3. & 4.
For question 1. this should be [here](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#83-filtering). 

As these questions are not specific to this API I suggest to park them to block the merging.

code/Test_definitions/sim-swap-subscriptions.feature
bigludo7: Yes this 'valid' wording is not defined and could be confusing as we are not supposed to store/manage _not valid_ subscription. I removed this unnecessary valid wording & make some ajustements in the wording.

code/Test_definitions/sim-swap-subscriptions.feature
```diff
@@ -306,39 +331,24 @@ Feature: CAMARA sim swap subscriptions  API, v0.1.1
 # Error Code 403
 ##################
 
-  @sim_swap_subscription_creation_60_phone_number_token_mismatch
+  @sim_swap_subscription_retrieval_61_phone_number_token_mismatch
```
fernandopradocabrillo, line 334: @bigludo7 I checked with Pedro regarding the SUBSCRIPTION_MISMATCH error because I wasn't fully sure if were right about the meaning. And spoiler: we weren't.
So this error is designed to cover the scenario where an API Client requests the creation of an event subscription for an event type for which it does not have permission. The token is valid and the client can consume the endpoint, but only for the events that were requested in the authentication.

Maybe with an example for device status this is clearer. There we have a scope for roaming on and another one for roaming off. If the token was generated for roaming on and the client requests the creation of a subscription for roaming off, there we should return the SUBSCRIPTION_MISMATCH error.

And this is why it is only present in the POST for creation

bigludo7: @fernandopradocabrillo Whaou got it ! now I remind ! - Kudos to Pedro
So I remove the tests for GET & DELETE
I'm not sure we need a Test for POST as we have only one event type.

fernandopradocabrillo (approved):

PR #180, opened 2024-12-23
Sim swap alignement with  commonalities 0.5

### LINKED ISSUE #161: Scope for Spring25 release (in preparation)

bigludo7:
#### What type of PR is this?

* enhancement/feature

Alignement of sim_swap.yaml, sim_swap-retrieveDate.feature and sim_swap-check.feature  with commonalities 0.5 rules.
See detail in release note

Fixes #161 

This is breaking change.

#### Changelog input

```
release-note
sim_swap.yaml:
- Change error structure definition to normalize error & status
- Update error code to introduce 429
- Remove 5xx errors
- Change in 422 code
- Add 3-legged token examples
- Improve documentation part

sim_swap-retrieveDate.feature:
- update test case scenario for 422
- Add optional test case for 401 AUTHENTICATION_REQUIRED

sim_swap-check.feature:
- update test case scenario for 422
- Add optional test case for 401 AUTHENTICATION_REQUIRED
```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -402,7 +402,7 @@ components:
                 code: SERVICE_NOT_APPLICABLE
                 message: The service is not available for the provided identifier.
             GENERIC_422_MISSING_IDENTIFIER:
-              description: An identifier is not included in the request and the device or phone number identification cannot be derived from the 3-legged access token
+              description: phone number is not included in the request (in case of 2-legged) or the phone number identification cannot be derived from access token (in 3-legged)
               value:
                 status: 422
                 code: MISSING_IDENTIFIER
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -469,7 +469,7 @@ components:
     CHECK_2LEGS:
       summary: Check request without 3-legged access tokens
       value:
-        phoneNumber: +346661113334
+        phoneNumber: "+346661113334"
         maxAge: 120
     CHECK_3LEGS:
       summary: Check request with 3-legged access tokens
@@ -478,7 +478,7 @@ components:
     RETRIEVE_2LEGS:
       summary: Retrieve request without 3-legged access tokens
       value:
-        phoneNumber: +346661113334
+        phoneNumber: "+346661113334"
     RETRIEVE_3LEGS:
       summary: Retrieve request with 3-legged access tokens
       value:
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -46,9 +46,9 @@ info:
 
     # Identifying the phone number from the access token
 
-    This API requires the API consumer to identify a device (via its phone number) as the subject of the API as follows:
-    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `phoneNumber` identifier, which therefore MUST be provided.
-    - When a three-legged access token is used however, this optional `phoneNumber` identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
+    This API requires the API consumer to identify a phone number as the subject of the API. There is 2 ways to retrieve it depending on the authorization flow used:
+    - When the API is invoked using a two-legged access token, the phone number will be identified from the optional `phoneNumber` identifier, which therefore MUST be provided.
+    - When a three-legged access token is used however, this optional `phoneNumber` identifier MUST NOT be provided, as the phone number will be uniquely identified from the access token.
 
     This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.
 
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -46,7 +46,7 @@ info:
 
     # Identifying the device from the access token
 
-    This API requires the API consumer to identify a device as the subject of the API as follows:
+    This API requires the API consumer to identify a device (via its phone number) as the subject of the API as follows:
     - When the API is invoked using a two-legged access token, the subject will be identified from the optional `phoneNumber` identifier, which therefore MUST be provided.
     - When a three-legged access token is used however, this optional `phoneNumber` identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
 
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -56,7 +56,7 @@ info:
 
     - If the subject cannot be identified from the access token and the optional `phoneNumber` identifier is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.
 
-    - If the subject can be identified from the access token and the optional `phoneNumber` identifier is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.
+    - If the phoneNumber can be identified from the access token and the optional `phoneNumber` identifier is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.
 
     # Further info and support
 
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -44,7 +44,7 @@ info:
 
     It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.
 
-    # Identifying the device from the access token
+    # Identifying the phone number from the access token
 
     This API requires the API consumer to identify a device (via its phone number) as the subject of the API as follows:
     - When the API is invoked using a two-legged access token, the subject will be identified from the optional `phoneNumber` identifier, which therefore MUST be provided.
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -40,33 +40,23 @@ info:
 
     The "Camara Security and Interoperability Profile" provides details on how a client requests an access token. Please refer to Identify and Consent Management (https://github.com/camaraproject/IdentityAndConsentManagement/) for the released version of the Profile.
 
-    Which specific authorization flows are to be used will be determined during onboarding process, happening between the API Client and the Telco Operator exposing the API, taking into account the declared purpose for accessing the API, while also being subject to the prevailing legal framework dictated by local legislation.
+    Which specific authorization flows are to be used will be determined during onboarding process, happening between the API Client and the API Provider, taking into account the declared purpose for accessing the API, while also being subject to the prevailing legal framework dictated by local legislation.
 
     It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.
 
-    # Identifying a phone number from the access token
+    # Identifying the device from the access token
 
-    This specification defines the `phoneNumber` field as optional in API requests, specifically in cases where the API is accessed using a 3-legged access token, and the phone number can be uniquely identified by the token. This approach simplifies API usage for API consumers by relying on the information associated with the access token used to invoke the API.
+    This API requires the API consumer to identify a device as the subject of the API as follows:
+    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `phoneNumber` identifier, which therefore MUST be provided.
+    - When a three-legged access token is used however, this optional `phoneNumber` identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
 
-    ## Handling of phone number information:
+    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.
 
-    ### Optional `phoneNumber` field for 3-legged tokens:
+    ## Error handling:
 
-    - When using a 3-legged access token, the phone number associated with the access token must be considered as the phone number for the API request. This means that the `phoneNumber` field is not required in the request, and if included it must identify the same phone number, therefore **it is recommended NOT to include it in these scenarios** to simplify the API usage and avoid additional validations.
+    - If the subject cannot be identified from the access token and the optional `phoneNumber` identifier is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.
 
-    ### Validation mechanism:
-
-    - The server will extract the phone number identification from the access token, if available.
-    - If the API request additionally includes a `phoneNumber` field when using a 3-legged access token, the API will validate that the phone number provided matches the one associated with the access token.
-    - If there is a mismatch, the API will respond with a 403 - INVALID_TOKEN_CONTEXT error, indicating that the phone number information in the request does not match the token.
-
-    ### Error handling for unidentifiable phone number:
-
-    - If the `phoneNumber` field is not included in the request and the phone number information cannot be derived from the 3-legged access token, the server will return a 422 `UNIDENTIFIABLE_PHONE_NUMBER` error.
-
-    ### Restrictions for tokens without an associated authenticated phone number:
-
-    - For scenarios which do not have a phone number associated to the token during the authentication flow, e.g. 2-legged access tokens, the `phoneNumber` field MUST be provided in the API request. This ensures that the phone number is explicit and valid for each API call made with these tokens.
+    - If the subject can be identified from the access token and the optional `phoneNumber` identifier is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.
 
     # Further info and support
 
@@ -78,7 +68,7 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
   version: wip
-  x-camara-commonalities: 0.4.0
+  x-camara-commonalities: 0.5.0
 externalDocs:
   description: Product documentation at Camara
   url: https://github.com/camaraproject/SimSwap
@@ -109,6 +99,12 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/CreateSimSwapDate"
+            examples:
+              RETRIEVE_3LEGS:
+                $ref: "#/components/examples/RETRIEVE_3LEGS"
+              RETRIEVE_2LEGS:
+                $ref: "#/components/examples/RETRIEVE_2LEGS"
+
         required: true
       responses:
         "200":
@@ -137,12 +133,8 @@ paths:
           $ref: "#/components/responses/Generic404"
         "422":
           $ref: "#/components/responses/Generic422"
-        "500":
-          $ref: "#/components/responses/Generic500"
-        "503":
-          $ref: "#/components/responses/Generic503"
-        "504":
-          $ref: "#/components/responses/Generic504"
+        "429":
+          $ref: "#/components/responses/Generic429"
   /check:
     post:
       security:
@@ -163,6 +155,11 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/CreateCheckSimSwap"
+            examples:
+              CHECK_3LEGS:
+                $ref: "#/components/examples/CHECK_3LEGS"
+              CHECK_2LEGS:
+                $ref: "#/components/examples/CHECK_2LEGS"
         required: true
       responses:
         "200":
@@ -184,12 +181,8 @@ paths:
           $ref: "#/components/responses/Generic404"
         "422":
           $ref: "#/components/responses/Generic422"
-        "500":
-          $ref: "#/components/responses/Generic500"
-        "503":
-          $ref: "#/components/responses/Generic503"
-        "504":
-          $ref: "#/components/responses/Generic504"
+        "429":
+          $ref: "#/components/responses/Generic429"
 components:
   securitySchemes:
     openId:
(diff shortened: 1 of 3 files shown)


code/API_definitions/sim-swap.yaml
```diff
@@ -40,33 +40,23 @@ info:
...
 
     It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.
 
-    # Identifying a phone number from the access token
+    # Identifying the device from the access token
 
-    This specification defines the `phoneNumber` field as optional in API requests, specifically in cases where the API is accessed using a 3-legged access token, and the phone number can be uniquely identified by the token. This approach simplifies API usage for API consumers by relying on the information associated with the access token used to invoke the API.
+    This API requires the API consumer to identify a device as the subject of the API as follows:
```
gregory1g, line 49: For SimSwap a subject of the API is a phone number, not a device. The "device-centric" language is misleading.

code/API_definitions/sim-swap.yaml
```diff
@@ -326,7 +354,7 @@ components:
               value:
                 status: 403
                 code: INVALID_TOKEN_CONTEXT
-                message: phoneNumber is not consistent with access token
+                message: "{{field}} is not consistent with access token."
```
gregory1g, line 357: What "field" other than the phoneNumber can it be?

code/API_definitions/sim-swap.yaml
```diff
@@ -427,3 +466,20 @@ components:
       value:
         latestSimChange: null
         monitoredPeriod: 120
+    CHECK_2LEGS:
+      summary: Check request without 3-legged access tokens
+      value:
+        phoneNumber: "+346661113334"
```
gregory1g, line 472: Is there any reason to keep phone number in the response at all?
This is a poll API - the invoker sent phoneNumber in the request, there is no value to send it back in the response.
Suggestion remove it from the repsonse completely.

From the backward compatibility point of view this is a gray area anyway.

code/API_definitions/sim-swap.yaml
```diff
@@ -40,33 +40,23 @@ info:
...
-    - If there is a mismatch, the API will respond with a 403 - INVALID_TOKEN_CONTEXT error, indicating that the phone number information in the request does not match the token.
-
-    ### Error handling for unidentifiable phone number:
-
-    - If the `phoneNumber` field is not included in the request and the phone number information cannot be derived from the 3-legged access token, the server will return a 422 `UNIDENTIFIABLE_PHONE_NUMBER` error.
-
-    ### Restrictions for tokens without an associated authenticated phone number:
-
-    - For scenarios which do not have a phone number associated to the token during the authentication flow, e.g. 2-legged access tokens, the `phoneNumber` field MUST be provided in the API request. This ensures that the phone number is explicit and valid for each API call made with these tokens.
+    - If the subject can be identified from the access token and the optional `phoneNumber` identifier is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.
```
gregory1g, line 59: I agree with the logic of the change, but this is a not-backward compatible change. Currently one may send 3-legged token and the phoneNumber (even if this is not recommended),  with this change this will cause a error.

code/API_definitions/sim-swap.yaml
```diff
@@ -109,6 +99,12 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/CreateSimSwapDate"
+            examples:
```
gregory1g, line 102: I did not found any formal description saying that phoneNumber must not be included in the response when 3-leged token is used. If such a rule is introduced, it is must be clearly described.
Many consider removing an optional response parameter as a not backward compatible change.

gregory1g (changes requested):

code/API_definitions/sim-swap.yaml
bigludo7: I do not follow you @gregory1g - in previous version the phoneNumber was **not** in the response.

code/API_definitions/sim-swap.yaml
bigludo7: Agree we have a breaking change here but we have to follow commonalities. Indeed, If you use 3-legs access token with a phoneNumber in the body you'll have a 422 UNNECESSARY_IDENTIFIER while previously it worked.

As a reminder we introduced this to avoid 'hidden' number verification

code/API_definitions/sim-swap.yaml
bigludo7: Humm...we do not the phoneNumber in the response - only in the request.

code/API_definitions/sim-swap.yaml
bigludo7: Agreed. I will apply.

code/API_definitions/sim-swap.yaml
bigludo7: I followed what is described in the [CAMARA_common.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml) 

I tend to think that this error should not be triggered in our case because if you pass a 3-legs access token identifying the phone number and a phone number in the body the 422 UNNECESSARY_IDENTIFIER must be triggered.

code/API_definitions/sim-swap.yaml
gregory1g: You are right. It looks like I checked wrong swagger file.
Still, is phone number really needed in the response in case of 2-legged? What for? The invoker provided the phoneNumber it is after in the request already, the response must deliver information about the requested phoneNumber. And if phone number is really needed in the successful response to re-check request/response ,match, why it is not included in all error responses as well?

code/API_definitions/sim-swap.yaml
gregory1g: The purpose of my comment was to highlight this, and make sure that the consequences of new commonalities are not overlooked.

code/API_definitions/sim-swap.yaml
gregory1g: You are right. something went wrong on my side here. ignore my comment.

code/API_definitions/sim-swap.yaml
gregory1g: Yes, I understood that you follow commonalities. But should we really copy paste its wording 1:1 when we know better. A developer will read this text and can be confused.

And you are right - with the change this PR introduces, this check will not be executed: 422 to be returned without checking if both match or not. Let's remove it.

bigludo7: Yes make sense for me. to remove ut but I would like to hear from @fernandopradocabrillo before to proceed.

Fernando do you think we can remove 403 INVALID_TOKEN_CONTEXT ? as we do not see when this error will be triggered.

bigludo7:
@fernandopradocabrillo @gregory1g  Following https://github.com/camaraproject/Commonalities/issues/369 discussion I think it's fair to remove 403 INVALID_TOKEN_CONTEXT example in the yaml.

May I ask you to review this new proposal soon as we have to prepare the RC candidate before Jan17th. We can still modify after but we have to achieve the milestone.

code/API_definitions/sim-swap.yaml
```diff
@@ -40,33 +40,23 @@ info:
...
 
-    Which specific authorization flows are to be used will be determined during onboarding process, happening between the API Client and the Telco Operator exposing the API, taking into account the declared purpose for accessing the API, while also being subject to the prevailing legal framework dictated by local legislation.
+    Which specific authorization flows are to be used will be determined during onboarding process, happening between the API Client and the API Provider, taking into account the declared purpose for accessing the API, while also being subject to the prevailing legal framework dictated by local legislation.
 
     It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.
 
-    # Identifying a phone number from the access token
+    # Identifying the device from the access token
```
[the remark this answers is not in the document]
bigludo7: You're right. I fixed line 49 but missed this one. I will fix. Thanks

code/API_definitions/sim-swap.yaml
```diff
@@ -351,68 +382,69 @@ components:
...
-      content:
-        application/json:
-          schema:
-            $ref: "#/components/schemas/ErrorInfo"
-          examples:
-            GENERIC_504_TIMEOUT:
-              description: API Server Timeout
+                status: 429
+                code: QUOTA_EXCEEDED
+                message: Either out of resource quota or reaching rate limiting.
```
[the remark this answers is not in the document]
bigludo7: This is a good comment @gregory1g but it should be triggered to commonalities first as here I used what we have defined for commonalities v0.5 (in https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml)

I suggest to not change this for this release to keep alignement with Commonalities 0.5 and then change this depending of commonalities outcome for next iteration. Work for you?

code/API_definitions/sim-swap.yaml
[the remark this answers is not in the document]
bigludo7: done

code/API_definitions/sim-swap.yaml
```diff
@@ -351,68 +382,69 @@ components:
...
               value:
                 status: 422
-                code: NOT_SUPPORTED
-                message: Service not supported for this phoneNumber
-            UNIDENTIFIABLE_PHONE_NUMBER:
-              description: The phone number is not included in the request and the phone number information cannot be derived from the 3-legged access token
+                code: SERVICE_NOT_APPLICABLE
+                message: The service is not available for the provided identifier.
+            GENERIC_422_MISSING_IDENTIFIER:
+              description: An identifier is not included in the request and the device or phone number identification cannot be derived from the 3-legged access token
```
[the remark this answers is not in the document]
bigludo7: This is generic.
I followed commonalities guideline see [here](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#61-standardized-use-of-camara-error-responses).

I suggest @gregory1g that you post your comment in commonalities as it makes sense.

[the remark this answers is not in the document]
gregory1g: created an issue there: https://github.com/camaraproject/Commonalities/issues/370

code/API_definitions/sim-swap.yaml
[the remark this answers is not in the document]
gregory1g: ok. Added https://github.com/camaraproject/Commonalities/issues/372

code/API_definitions/sim-swap.yaml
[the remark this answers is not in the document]
gregory1g: according to Eric, commonality's "description" is not a error message to be send, it is a description for the reader of the table in the document. While example of the message is in the "Message example" column. An this is just an example, different API can use different text

code/API_definitions/sim-swap.yaml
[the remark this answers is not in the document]
gregory1g: according to Eric "Message example" column where the text came from us just an example, different API can use different text. So, there is no need to use exactly the same text as in the commonalities.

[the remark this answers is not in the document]
bigludo7: Thanks @gregory1g 
As the implementation is free to change the `description` do we have to provide here something different from other API for 429?
I 'm fine to provide 2 exemple if is it a blocking point for you but me preference is to keep as it.
+ @fernandopradocabrillo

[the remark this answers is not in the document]
bigludo7: Done

code/API_definitions/sim-swap.yaml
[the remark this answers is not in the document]
bigludo7: I've reworked the yaml & TC to add all clarification - please review.

code/API_definitions/sim-swap.yaml
```diff
@@ -40,33 +40,23 @@ info:
...
 
     It is important to remark that in cases where personal user data is processed by the API, and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of 3-legged access tokens becomes mandatory. This measure ensures that the API remains in strict compliance with user privacy preferences and regulatory obligations, upholding the principles of transparency and user-centric data control.
 
-    # Identifying a phone number from the access token
+    # Identifying the phone number from the access token
 
-    This specification defines the `phoneNumber` field as optional in API requests, specifically in cases where the API is accessed using a 3-legged access token, and the phone number can be uniquely identified by the token. This approach simplifies API usage for API consumers by relying on the information associated with the access token used to invoke the API.
+    This API requires the API consumer to identify a device (via its phone number) as the subject of the API as follows:
```
gregory1g, line 49: "to identify a device" this is not the case, the API does not need to identify a device at all. MSIDN is subscription ID, not a device ID. And the API implementation does not need to identify the device either.
Considering that API description targets application developers, using "device" is rather misleading. It makes a reader think that this API target exact phone unit like geo location APIs do.

code/API_definitions/sim-swap.yaml
```diff
@@ -40,33 +40,23 @@ info:
...
+    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `phoneNumber` identifier, which therefore MUST be provided.
```
gregory1g, line 50: "the subject will be identified from the optional `phoneNumber` identifier" says that the phoneNumner is not the subject of the API, but just a helper to identify the device. This is confusing. The original terminology was simpler to understand. Especially when one also reads documentation for other API which really target a device.

code/API_definitions/sim-swap.yaml
```diff
@@ -40,33 +40,23 @@ info:
...
+    - When a three-legged access token is used however, this optional `phoneNumber` identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
 
-    ## Handling of phone number information:
+    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.
 
-    ### Optional `phoneNumber` field for 3-legged tokens:
+    ## Error handling:
 
-    - When using a 3-legged access token, the phone number associated with the access token must be considered as the phone number for the API request. This means that the `phoneNumber` field is not required in the request, and if included it must identify the same phone number, therefore **it is recommended NOT to include it in these scenarios** to simplify the API usage and avoid additional validations.
+    - If the subject cannot be identified from the access token and the optional `phoneNumber` identifier is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.
```
gregory1g, line 57: Suggestion: use "phoneNumer" instead of "subject" - it is still a single word and it is simpler to understand.

code/API_definitions/sim-swap.yaml
bigludo7: Fixed accordingly

code/API_definitions/sim-swap.yaml
bigludo7: Fixed accordingly

code/API_definitions/sim-swap.yaml
```diff
@@ -40,33 +40,23 @@ info:
...
+    - When a three-legged access token is used however, this optional `phoneNumber` identifier MUST NOT be provided, as the phone number will be uniquely identified from the access token.
 
-    ## Handling of phone number information:
+    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.
 
-    ### Optional `phoneNumber` field for 3-legged tokens:
+    ## Error handling:
 
-    - When using a 3-legged access token, the phone number associated with the access token must be considered as the phone number for the API request. This means that the `phoneNumber` field is not required in the request, and if included it must identify the same phone number, therefore **it is recommended NOT to include it in these scenarios** to simplify the API usage and avoid additional validations.
+    - If the subject cannot be identified from the access token and the optional `phoneNumber` identifier is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.
```
fernandopradocabrillo, line 57:
```suggestion
    - If the phoneNumber cannot be identified from the access token and the optional `phoneNumber` identifier is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.
```

code/API_definitions/sim-swap.yaml
```diff
@@ -40,33 +40,23 @@ info:
...
-    - If there is a mismatch, the API will respond with a 403 - INVALID_TOKEN_CONTEXT error, indicating that the phone number information in the request does not match the token.
-
-    ### Error handling for unidentifiable phone number:
-
-    - If the `phoneNumber` field is not included in the request and the phone number information cannot be derived from the 3-legged access token, the server will return a 422 `UNIDENTIFIABLE_PHONE_NUMBER` error.
-
-    ### Restrictions for tokens without an associated authenticated phone number:
-
-    - For scenarios which do not have a phone number associated to the token during the authentication flow, e.g. 2-legged access tokens, the `phoneNumber` field MUST be provided in the API request. This ensures that the phone number is explicit and valid for each API call made with these tokens.
+    - If the subject can be identified from the access token and the optional `phoneNumber` identifier is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.
```
fernandopradocabrillo, line 59:
```suggestion
    - If the phoneNumber can be identified from the access token and the optional `phoneNumber` identifier is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.
```

fernandopradocabrillo (dismissed):
LGTM

code/API_definitions/sim-swap.yaml
eric-murray: > using "device" is rather misleading

One of the advantages of the CAMARA approach is that application developers can use APIs such as SIM Swap without knowing (or learning) the MSISDN at all. All they need to do is identify a device using the subscription, and capture its IP address. I think ruthlessly expunging "device" from our documentation and baking in "phone number" mean this possibility gets overlooked.

It would be nice if we could come up with some "developer friendly" formulation of words which also captures this possibility, as there will be use cases out there that would benefit from this approach if only developers were aware of it.

bigludo7: @eric-murray To be sure to follow you - as in the body we can only use the phone number  - you want to highlight the point that via the 3-legged token we can value in the `login-hint` other identifier like ip?

eric-murray: @bigludo7
No, it was just a general comment - I'm not expecting the documentation to be updated because of it.

Using a 2-legged access token and explicit identifier will very much be the exception for CAMARA APIs, and yet it is the definition of the explicit identifier that dominates the API description. For the normal case of a 3-legged access token, the token itself is not classified as a "device" token or "phone number" token. The very same token could be used for both a "phone number" API (such as SIM Swap) and a "device" API (such as Device Status).

If 3-legged tokens were the only option, we would not have to go through these mental gymnastics of documenting an API as either a "device" API or a "phone number" API. It is to be hoped that we can come to some agreement on a common formulation for the documentation that reveals this capability to developers without confusing them.

gregory1g: @eric-murray 
> "All they need to do is identify a device using the subscription, and capture its IP address." 

not exactly, SimSwap API answers the question "if a SIM Card was swapped for the MSISDN". MSISDN is a target of the API not a device. SimSwap can potentially processes an IP as an input, but then it will need to find MSISDN associated with this device first and then do the checks for this MSISDN. For example, if an additional SIM was issued for the given MSISND and used in completely different device, it is still a SimSwap.

fernandopradocabrillo (approved):

code/API_definitions/sim-swap.yaml
eric-murray: > SimSwap can potentially processes an IP as an input

For a 3-legged access token, the "processing" will already have been done. The SIM Swap API itself does not need to do anything different.

I take the point that IP as a device identifier is not persistent. An API consumer cannot capture the IP address and expect to use that the following week to check the SIM Swap status of the associated subscription. But if they immediately obtain an access token for that IP, then they can check the SIM Swap status for the subscription used by that device, all without even knowing what the MSISDN is.

It is this possibility I think gets overlooked - using "phone number" APIs without knowing the phone number.

gregory1g: my point was that SimSwap does not target a device, it targets an MSISDN.

bigludo7: @gregory1g @eric-murray Does this discussion prevent me to merge this PR? with the M3 at the horizon we need to move forward for the yaml & work on the other asset. I can try to copy paste your comment in the discussion?

eric-murray: @bigludo7
No changes needed, just proceed with the M3 as planned.

Probably the full range of use cases that can be supported by an API can be better advertised through developer portals rather than the YAML documentation anyway. But we can maybe think later on how we can better document what a developer needs to know or needs access to in order to use an API.

gregory1g:
lgtm

gregory1g (approved):

PR #189, opened 2025-02-25
Add 400 OUT_OF_RANGE error when the maxAge is above 2400

### LINKED ISSUE #187: Add 400 with OUT_OF_RANGE code in sim-swap 2.0.0
**Problem description**
Following discussion #184 suggestion to add 400 with OUT_OF_RANGE code in sim-swap 2.0.0

**Possible evolution**

- Add in the yaml this code
- Add a test case

**Alternative solution**

**Additional context**
I suggest to add this one in the current release as we have a major version.
@hdamker @fernandopradocabrillo @QaunainM WDYT?

bigludo7:
#### What type of PR is this?

* enhancement/feature

- Add 400 OUT_OF_RANGE error when the maxAge is above 2400

Fixes #187 

#### Changelog input

```
 release-note
- Add 403 with code OUT_OF_RANGE when the maxAge is above 2400


```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -288,6 +288,7 @@ components:
                   code:
                     enum:
                       - INVALID_ARGUMENT
+                      - OUT_OF_RANGE
           examples:
             GENERIC_400_INVALID_ARGUMENT:
               description: Invalid Argument. Generic Syntax Exception
@@ -295,6 +296,12 @@ components:
                 status: 400
                 code: INVALID_ARGUMENT
                 message: Client specified an invalid argument, request body or query param.
+            GENERIC_400_OUT_OF_RANGE:
+              description: Out of Range. Specific Syntax Exception used when a given field has a pre-defined range or a invalid filter criteria combination is requested
+              value:
+                status: 400
+                code: OUT_OF_RANGE
+                message: Client specified an invalid range.
     Generic401:
       description: Unauthorized
       headers:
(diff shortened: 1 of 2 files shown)


hdamker:
@bigludo7 I edited the title and the description of the PR, as the content is a 400, not a 403 error response.

code/API_definitions/sim-swap.yaml
```diff
@@ -288,13 +288,20 @@ components:
...
               value:
                 status: 400
                 code: INVALID_ARGUMENT
                 message: Client specified an invalid argument, request body or query param.
+            GENERIC_400_OUT_OF_RANGE:
+              description: Out of Range. Specific Syntax Exception used when a given field has a pre-defined range or a invalid filter criteria combination is requested
+              value:
+                status: 400
+                code: OUT_OF_RANGE
+                message: Client specified an invalid range.
```
gregory1g, line 304: This is a very generic definition. A client which will receive such error code will have to guess what exactly value is out of the range and what this range is. The error must explicitly say this.
If we introduce this error for maxAge > 24000 only - we can describe this in he API description. So, a client knows what the error is really about.

bigludo7: As we have very few attribute in this check request (only the maxAge in 3-legs, maxAge & phoneNumber in the 2-legs scenario) I have no problem to custom our message to MAX_AGE_OUT_OF_RANGE.
@gregory1g @fernandopradocabrillo @QaunainM work for you?

fernandopradocabrillo: I prefer to stick with the generic one from commonalities artifact. The message is a free text field that can be customized by the implementation. If you feel that it is something critical we can use a message like: "Client specified a value of maxAge outside the defined range"

bigludo7: @fernandopradocabrillo Yes make sense - I've forget our message capability - This seems to me a better approach.

fernandopradocabrillo: To have in mind for this PR: https://github.com/camaraproject/SimSwap/discussions/184#discussioncomment-12336451

PR #200, opened 2025-03-26
Include 403 INVALID_TOKEN_CONTEXT and SUBSCRIPTION_DELETED terminationReason

### LINKED ISSUE #196: SUBSCRIPTION_DELETED event and INVALID_TOKEN_CONTEXT error missing in Fall24 v0.1.1 sim-swap-subscriptions
**Problem description**

In the sim-swap-subscriptions API there are two mismatches with Commonalities that leave gaps in implementation:
1. The error `403 INVALID_TOKEN_CONTEXT` is missing in the POST that creates the subscription. This error is needed since in this version we still allow to send a phoneNumber in the request body in a 3-legged scenario and we should be able to validate it.
2. The event `SUBSCRIPTION_DELETED` is missing -> This is a bug since we have no clear way to inform of the cancelation of a subscription by the API Client.

**Expected behavior**

From TEF we would like to update the yaml to v0.1.2 and create a new maintenance release to fix these bugs so the Fall24 version is usable.

fernandopradocabrillo:
#### What type of PR is this?

* bug
* correction

This PR includes two fixes into Fall24 sim swap subscriptions API:
1. `HTTP 403 INVALID_TOKEN_CONTEXT error`
2. `SUBSCRIPTION_DELETED` terminationReason for `SUBSCRIPTION_ENDS` event
3. Include subscription for this terminationReason

Fixes #196 

This changes will be included in v0.1.2 of sim-swap-subscriptions which is the version belonging to the Fall24 meta-release

### DIFF
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -35,6 +35,7 @@ info:
     It is used when:
       - the subscription expire time (optionally set by the requester) has been reached
       - the maximum number of subscription events (optionally set by the requester) has been reached
+      - the subscription was deleted by the requester
       - the API server has to stop sending notification prematurely
       - the Access Token sinkCredential (optionally set by the requester) expiration time has been reached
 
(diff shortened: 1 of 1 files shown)


bigludo7 (dismissed):
LGTM

hdamker (changes requested):
This PR has the wrong merge target if you want to create a patch of r1.3. It must be camaraproject:maintenance-r1 (to be created as first step before the PR in the camaraproject repository).

See also https://github.com/camaraproject/SimSwap/issues/196#issuecomment-2755160683

hdamker:
Now it looks right to me :-)

bigludo7 (approved):
LGTM

PR #206, opened 2025-05-07
[Sim-Swap-Subscriptions]: Remove `phoneNumber` from the required properties in the Swap-Event

### LINKED ISSUE #197: `data.phoneNumber` cannot be required in notifications for sim-swap-subscriptions
**Problem description**
Currently `data.phoneNumber` is required in notifications for sim-swap-subscriptions. But if the subscription was created using a 3-legged access token, then this would allow the API consumer to match the pseudonymised `sub` field of the ID token with the subscriber's true identity.

In some scenarios (e.g. if the 3-legged access token was created using the authorisation code flow, or using an operator token), the API consumer would not necessarily know, and may not be allowed to know, the true identity of the mobile subscriber.

**Expected behavior**
`phoneNumber` should be optional in the `Swapped` schema, and only included if a 2-legged access token was used to create the subscription in the first place. Note also the update in the `description`.
```
    Swapped:
      description: Event detail structure for a SWAPPED event
      type: object
      required:
        - subscriptionId
      properties:
        phoneNumber:
          $ref: '#/components/schemas/PhoneNumber'
        subscriptionId:
          $ref: "#/components/schemas/SubscriptionId"
```
**Alternative solution**

**Additional context**

maxl2287:
#### What type of PR is this?

* correction

The `phoneNumber` is optional in the Event, when 3-legged-token was used.

Fixes #197

### DIFF
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -712,7 +712,6 @@ components:
       description: Event detail structure for ROAMING_ON & ROAMING_OFF event
       type: object
       required:
-        - phoneNumber
         - subscriptionId
       properties:
         phoneNumber:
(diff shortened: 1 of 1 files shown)


fernandopradocabrillo (approved):

bigludo7 (approved):
LGTM

PR #209, opened 2025-05-08
[Sim-Swap-Subscriptions]: remove `SUBSCRIPTION_MISMATCH` error-code

### LINKED ISSUE #194: 403 SUBSCRIPTION_MISMATCH cannot be returned
**Problem description**

In the subscriptions API, the error 403 SUBSCRIPTION_MISMATCH of the Create operation is returned when the access token has been granted with a scope that allows to create a subscription for an event but, in the request body, the client requests a subscription for a different one.
This is fine in APIs where there is more that one event to subscribe to but here, since we only have one, if the token lacks the scope, the error returned would be 403 PERMISSION_DENIED instead.

**Expected behavior**

Remove error 403 SUIBSCRIPTION_MISMATCH

**Alternative solution**

Open to suggestions

maxl2287:
#### What type of PR is this?

* correction

Removed `SUBSCRIPTION_MISMATCH` error as it is not applicable when only one event type can be subscribed to.

Fixes #194 

#### Changelog input

```
 release-note
 * [Sim-Swap-Subscriptions]: remove `SUBSCRIPTION_MISMATCH` error-code
```

### DIFF
--- a/code/API_definitions/sim-swap-subscriptions.yaml
+++ b/code/API_definitions/sim-swap-subscriptions.yaml
@@ -1085,7 +1085,6 @@ components:
                   code:
                     enum:
                       - PERMISSION_DENIED
-                      - SUBSCRIPTION_MISMATCH
           examples:
             GENERIC_403_PERMISSION_DENIED:
               description: Permission denied. OAuth2 token access does not have the required scope or when the user fails operational security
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
LGTM
Thanks Max

fernandopradocabrillo (approved):

PR #210, opened 2025-05-14
Add maxl2287 as Codeowner

maxl2287:
#### What type of PR is this?

* subproject management

After talking to @bigludo7 and @hdamker, I would like to support here with 2 more eyes in SimSwap :)


bigludo7 (approved):
Fully support:)

hdamker (approved):

PR #250, opened 2026-02-11
[WIP] feat: add optional 'ageTiers' for granular risk assessment

### LINKED ISSUE #248: Optional ageTiers parameter for privacy-preserving granularity
**Problem description**
Current consumers of the API (e.g., banks) often need more granular signals to assess account-takeover risk alongside their own data. A simple boolean result (True/False) doesn’t provide enough detail, while returning the underlying event date/time isn’t feasible due to strict privacy requirements (e.g., GDPR) and operator restrictions. As a result, clients must make multiple calls with different maxAge values (for example, 1 hour, then 24 hours) to infer the risk level—creating unnecessary network and API load and poor overall performance.

**Possible evolution**
Introduce an optional input parameter ageTiers (list of integers) that enables a "Tiered Response" mode, where the highest tier is the maxAge.

Backward Compatible: If ageTiers is omitted, the API behaves exactly as it does today (using maxAge).

Privacy Friendly: Returns a tier index rather than an exact timestamp.

Efficient: Reduces N API calls to 1 call.

**Alternative solution**
make multiple calls with different maxAge values

**Additional context**
Samples1: Swapped at 3 hour ago

input:
{
  "phoneNumber": "+1234567890",
  "maxAge": 240,
  "ageTiers": [1,4,24,72,240]
}

output:
{
  "swapped": true, // Calculated using maxAge (3 < 240)
  "tierIndex": 1       // 3 hours is > 1 but <= 4. It falls in the second bucket (Index 1).
}

Samples1: Swapped at more than 240 hour ago
output:
{
  "swapped": false, 
  "tierIndex": -1       // out of the maxAge
}

yyeAduna:
Introduce an optional input parameter ageTiers (list of integers) that enables a "Tiered Response" mode, where the highest tier is the maxAge.

What type of PR is this?

enhancement/feature

What this PR does / why we need it:

Introduces an additive, optional mechanism to query SIM swap "recency tiers" without exposing exact timestamps.

Current consumers of the API (e.g., banks) often need more granular signals to assess account-takeover risk alongside their own data for their ML based risk engine (vs the rule based sequential gate). A simple boolean result (True/False) is form of lossy compression and doesn’t provide enough detail, while returning the underlying event date/time isn’t feasible due to strict privacy requirements (e.g., GDPR) and operator restrictions.

Also due to privacy and security reasons, a bank won't want to leak the value/risk of their customer's operation and disclose their internal risk policies to third party by using different maxAge for each individual transactions. The standard age-tier approach will help to protect bank customers' privacy and bank's corporate security.

Solution:

Adds ageTiers (Request): An optional list of integer buckets (e.g., [1, 4, 24]) with max 10 tiers

Adds tierIndex (Response): An optional integer indicating which bucket the swap falls into.

Which issue(s) this PR fixes:

Fixes https://github.com/camaraproject/SimSwap/issues/248

Special notes for reviewers:

Backward Compatibility Analysis: This change is strictly additive.

The maxAge parameter remains required and continues to drive the standard swapped boolean logic.

If ageTiers is omitted by the client (legacy behavior), the API behaves exactly as the current v2.0.0 spec, and tierIndex is not returned.

Privacy Impact: This improvement enhances privacy by allowing operators to share risk granularity without revealing the exact time of the swap (PII). Also allow the banks (Clients) take full advantage of the function without revealing the value/risk of customer's bank transactions/operations.

### DIFF
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -34,8 +34,9 @@ info:
 
       - If the latest SIM swap date (or the activation date if no SIM swap) cannot be communicated due to local regulations (or MNO internal privacy policies) preventing the safekeeping of the information for longer than the stated period, a `null` value will be returned. Optionally, a `monitoredPeriod` could be provided to indicate monitored time frame (in days) supported by the MNO. In this case the response must be treated as "there were no sim swap events during 'monitoredPeriod'. Either the parameter is optional, it is recommended to support it in SimSwap implementations.
 
-    - POST check: Checks if SIM swap has been performed during a past period (defined in the request with 'maxAge' attribute) for a given phone number.
-
+    - POST check: Checks if SIM swap has been performed during a past period for a given phone number.
+      - The API returns a boolean response (true/false) based on the mandatory 'maxAge' attribute.
+      - Optionally, the client can provide 'ageTiers' (list of time buckets) to receive a 'tierIndex', enabling granular risk assessment without exposing exact timestamps.
       - Note: In the specification of the API the 'maxAge' could be between 1 hour to 2400 hours. If this delay is not managed due to the operator's own privacy threshold (which in theory would likely be linked to local regulations in the country) and a request is performed with a value inferior to 2400 but superior to operator policy, then,  an error `400 OUT_OF_RANGE `is expected with an explicit message to explain the limitation like `Check monitor period could not exceed local regulations (30 days)`.
 
     # Authorization and authentication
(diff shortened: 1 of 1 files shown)


bigludo7:
Hi @yyeAduna 
Thanks !
May I ask you to trigger an issue to gather team feedback.

To be honest this is a breaking change (not from technical perspective but fonctional/business) and from my company we're not sure about the breaking change vs evolution value.
I understand the business requirement but in this case I will see this as a L3 api (composite API) offered by channel partner and consuming atomic L1 MNO API.

I tag some usual suspects to get their view: @fernandopradocabrillo @HuubAppelboom @jgarciahospital @AxelNennker

HuubAppelboom:
> Hi @yyeAduna Thanks ! May I ask you to trigger an issue to gather team feedback.
> 
> To be honest this is a breaking change (not from technical perspective but fonctional/business) and from my company we're not sure about the breaking change vs evolution value. I understand the business requirement but in this case I will see this as a L3 api (composite API) offered by channel partner and consuming atomic L1 MNO API.
> 
> I tag some usual suspects to get their view: @fernandopradocabrillo @HuubAppelboom @jgarciahospital @AxelNennker

This could get quite complex, better put this first in a separate issue or discussion topic

yyeAduna:
reopen issue https://github.com/camaraproject/SimSwap/issues/248 to facilitate discussion

bigludo7:
Replaced by #266

PR #254, opened 2026-03-19
Update sim-swap.yaml with age band

shilpa-padgaonkar:
#### What type of PR is this?

Add one of the following kinds:
* enhancement/feature

#### What this PR does / why we need it:
Extends the retrieve-date endpoint to respond with a standardized recency bucket representing the elapsed time between the SIM change date and the current time.

#### Which issue(s) this PR fixes:

Fixes #253 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

Issue #253, opened 2026-03-13
Optional standardized operator-defined recency bucket to be returned with swapped boolean(SimSwap)


KeldaAnders:
**Problem description**
Consumers have requested coarse time-band signals to support better fraud and risk decisioning. The current /check endpoint returns only a Boolean (swapped), which by itself is an insufficient signal.

---

## Possible evolution

Return an Optional **standardized operator-defined recency bucket**.

Proposed approach:

- Keep the existing **Boolean `swapped`** field.
- Add an optional **`simSwapAgeBandEnum`**.
- Preserve backward compatibility.

This keeps `/check` aligned with its purpose as a **simple threshold signal** while optionally exposing standardized context.

The consuming party (e.g., a bank) can then apply its own thresholds, ML features, or step-up logic **outside the API contract**.

---

### Request

```json
{
  "phoneNumber": "+1234567890"
}
```

The request remains intentionally minimal and **consumer-agnostic**.

---

---

### Response

```json
{
  "swapped": true,
  "simSwapAgeBandEnum": 1,
}
```

- `swapped` remains the **primary contract**
- enum fields provide **optional additional signal**

---

## YAML schema note

Suggested explanatory note:

> `simSwapAgeBandEnum` is an optional standardized recency bucket representing the elapsed time between the SIM activation/change date and the current time.  
> The value maps the duration into fixed bands (0–14), ranging from 0 hours up to 3+ years.

Special values:

- `111` → Change has not occurred  
- `999` → Subscriber valid but no real-time profile available  

---

## Standardized AgeBandEnum mapping

```
0  = 0h ≤ d < 4h
1  = 4h ≤ d < 12h
2  = 12h ≤ d < 1d
3  = 1d ≤ d < 2d
4  = 2d ≤ d < 5d
5  = 5d ≤ d < 7d
6  = 7d ≤ d < 14d
7  = 14d ≤ d < 30d
8  = 30d ≤ d < 60d
9  = 60d ≤ d < 90d
10 = 90d ≤ d < 180d
11 = 180d ≤ d < 1y
12 = 1y ≤ d < 2y
13 = 2y ≤ d < 3y
14 = 3y+

111 = Change has not happened
999 = Subscriber valid, but no real-time profile found
```

```
enum: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,111,999]
```

## Backward compatibility

This proposal introduces **additive-only changes**:

- Existing `swapped` behavior remains unchanged
[...]

PR #266, opened 2026-06-05
Update sim-swap | add /retrieve-age-band endpoint returning age band …

### LINKED ISSUE #260: Add a third SIM Swap API endpoint to return SIM-Swap age band
**Problem description**

The current `/check` response provides a Boolean `swapped` field, which is useful as a simple threshold signal but does not provide standardized context about the recency of the SIM activation/change event.

Consumers such as banks, wallets, and fraud platforms often need more than a binary `swapped / not swapped` response. Different consumers may apply different policies depending on whether a SIM change happened within hours, days, weeks, months, or years.

Without a standardized recency bucket, consumers may receive inconsistent interpretations across operators, aggregators, and implementations.

**Possible evolution**

Keep the existing Boolean `swapped` field as the primary contract and add an optional standardized `simSwapAgeBandEnum` response field.

The request remains intentionally minimal and consumer-agnostic:

```json
{
  "phoneNumber": "+1234567890"
}
````

Example response:

```json
{
  "swapped": true,
  "simSwapAgeBandEnum": 1
}
```

Proposed schema note:

`simSwapAgeBandEnum` is an optional standardized recency bucket representing the elapsed time between the SIM activation/change date and the current time. The value maps the duration into fixed bands, ranging from 0 hours up to 3+ years.

Proposed standardized `AgeBandEnum` mapping:

```text
-1 = Subscriber valid, but no real-time profile found
 0 = Change has not happened
 1 = 0h ≤ d < 4h
 2 = 4h ≤ d < 12h
 3 = 12h ≤ d < 1d
 4 = 1d ≤ d < 2d
 5 = 2d ≤ d < 5d
 6 = 5d ≤ d < 7d
 7 = 7d ≤ d < 14d
 8 = 14d ≤ d < 30d
 9 = 30d ≤ d < 60d
10 = 60d ≤ d < 90d
11 = 90d ≤ d < 180d
12 = 180d ≤ d < 1y
13 = 1y ≤ d < 2y
14 = 2y ≤ d < 3y
15 = 3y+
```

Suggested enum:

```yaml
enum: [-1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
```

The enum should be fully standardized, not operator-defined. A standardized enum provides consistent interpretation across operators, aggregators, and consuming applications.

Special states should be represented as enum values to keep the response compact and self-contained:

```text
-1 = Subscriber valid, but no real-time profile found
 0 = Change has not happened
```

Negative enum values should be acceptable for this API schema when used for clearly defined special cases. The `-1` value is useful because it distinguishes a valid subscriber with unavailable real-time profile data from both “no SIM change” and actual recency buckets.

**Additional context**

The `3y+` upper bucket provides useful SIM context and an extended stability signal.

A very long unchanged SIM profile can provide positive context to consuming parties such as banks, wallets, and fraud platforms. The `3y+` bucket is driven by customer demand. We have verifiable customer requests and traffic using this recency bucket.

Customers are asking not only for a Boolean SIM swap signal, but also for a standardized recency context that lets them tune their own risk thresholds and decisioning logic.

This is also based on observed fraud and risk decisioning needs:

* A SIM change within the last few hours may trigger a hard block, step-up authentication, or manual review.
* A SIM change within the last few days may increase risk scoring without automatically blocking the user.
* A SIM profile unchanged for multiple years may be treated as a positive stability signal.
* ML-based systems can use the standardized bucket as a feature without requiring the API contract itself to encode customer-specific thresholds.

Standardized age bands support both fraud detection and the continuity of legitimate users while keeping the API simple and consumer-agnostic.

KeldaAnders:
**Title:** feat: add `/retrieve-age-band` endpoint returning SIM swap age band only

Adds a third, single-purpose operation that returns the standardized `simSwapAgeBandEnum` only; excludes the `swapped` Boolean (`/check`) and the change date (`/retrieve-date`) per the 2026-04-29 ad-hoc agreement (#253) and TEF stance (#260). Targeting release **2.1.0**.

#### What type of PR is this?

> enhancement/feature

Adds `POST /retrieve-age-band` to the SIM Swap API, a third endpoint that returns a standardized recency band (`simSwapAgeBandEnum`) for the last SIM swap/activation event — and nothing else.

- **Purely additive.** `/check` and `/retrieve-date` (and their schemas, request bodies, and examples) are preserved unchanged. The only additions are the new path, the `SimSwapAgeBandEnum` / `SimSwapAgeBandInfo` / `CreateSimSwapAgeBand` schemas, the `AGEBAND_*` examples, and the description/scope updates.
- **Single-purpose by design.** The response carries the band value only. It does **not** include the `swapped` Boolean (that is `/check`) or the SIM change timestamp (that is `/retrieve-date`), keeping each endpoint contract clean — which is the scope agreed on the 2026-04-29 ad-hoc call (#253) and in TEF's position on #260, and which addresses the scope concern raised on #254.
- **Band mapping** retains the standardized values: `-1` = subscriber valid but no real-time profile, `0` = no change, `1`–`15` = recency buckets (4h → 3y+).
- **New scope** `sim-swap:retrieve-age-band`, alongside the existing `sim-swap:check` and `sim-swap:retrieve-date` and the API-level `sim-swap` scope.

Adding a new endpoint is a backward-compatible change, so this targets a MINOR bump to **2.1.0** (URL path unchanged: `/sim-swap/v2`).

- **Scope is the point.** Per #253/#260 this is deliberately a *third* endpoint exposing the age band alone — not an extension of `/check` or `/retrieve-date`. Please flag if any check/date logic has leaked in. Supersedes #254.
- **Additive, not a rewrite.** The diff should show **no deletions** in the `/check` or `/retrieve-date` blocks. If you see those as removed, the wrong (single-endpoint) file was committed.
- **Commonalities deferred on purpose.** This keeps the `-1` sentinel and `x-camara-commonalities: 0.6`. Aligning the age-band response to the newer Commonalities business-outcome pattern (`contextCode`/`contextMessage` for unknown/unavailable results) would require bumping Commonalities for the whole file and is intentionally **not** done here — raising it as a separate change. Happy to follow up with that bump if the WG prefers.
- **Co-located, not a separate API.** The age band is another read over the same underlying SIM-swap event, alongside `/check` and `/retrieve-date`, so it is added to `sim-swap.yaml` rather than split into an atomic API. Open to discussion.
- **Version field** stays `wip` on `main`; the release number (2.1.0) is assigned at release time.

#### Changelog input

```release-note
Added a new `POST /retrieve-age-band` endpoint returning the standardized SIM swap recency band (`simSwapAgeBandEnum`) only, without the `swapped` Boolean or change date. Existing `/check` and `/retrieve-date` endpoints are unchanged.
```

#### Additional documentation

This section can be blank.

```docs

```

### DIFF
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -2,7 +2,7 @@ openapi: 3.0.3
 info:
   title: SIM Swap
   description: |
-    The SIM swap API provides a programmable interface for developers and other users (capabilities consumers) to request the last date of a SIM swap performed on the mobile line, or, to check whether a SIM swap has been performed during a past period.
+    The SIM swap API provides a programmable interface for developers and other users (capabilities consumers) to request the last date of a SIM swap performed on the mobile line, to check whether a SIM swap has been performed during a past period, or to retrieve a standardized recency band for the last SIM swap event.
 
     # Introduction
 
@@ -12,10 +12,13 @@ info:
 
     The SIM Swap API can also be used to protect non-automated actions. For example, when a call center expert contacts a user to clarify or confirm a sensitive operation.
 
-    This API is used by an application to get information about a mobile line latest SIM swap date. It can be easily integrated and used through this secured API and allows SPs (Service Provider) to get this information an easy & secured way. The API provides management of 2 endpoints answering 2 distinct questions:
+    This API is used by an application to get information about a mobile line latest SIM swap date. It can be easily integrated and used through this secured API and allows SPs (Service Provider) to get this information an easy & secured way. The API provides management of 3 endpoints answering 3 distinct questions:
 
     * When did the last SIM swap occur?
+
     * Has a SIM swap occurred during last n hours?
+
+    * How long ago did the last SIM swap or activation occur (as a standardized recency band)?
 
     # Relevant terms and definitions
 
@@ -26,7 +29,7 @@ info:
 
     # API functionality
 
-    The API provides 2 operations:
+    The API provides 3 operations:
 
     - POST retrieve-date : Provides timestamp of latest SIM swap, if any, for a given phone number.
 
@@ -36,7 +39,28 @@ info:
 
     - POST check: Checks if SIM swap has been performed during a past period (defined in the request with 'maxAge' attribute) for a given phone number.
 
-      - Note: In the specification of the API the 'maxAge' could be between 1 hour to 2400 hours. If this delay is not managed due to the operator's own privacy threshold (which in theory would likely be linked to local regulations in the country) and a request is performed with a value inferior to 2400 but superior to operator policy, then,  an error `400 OUT_OF_RANGE `is expected with an explicit message to explain the limitation like `Check monitor period could not exceed local regulations (30 days)`.
+      - Note: In the specification of the API the 'maxAge' could be between 1 hour to 2400 hours. If this delay is not managed due to the operator's own privacy threshold (which in theory would likely be linked to local regulations in the country) and a request is performed with a value inferior to 2400 but superior to operator policy, then, an error `400 OUT_OF_RANGE `is expected with an explicit message to explain the limitation like `Check monitor period could not exceed local regulations (30 days)`.
+
+    - POST retrieve-age-band : Returns a standardized `simSwapAgeBandEnum` recency bucket representing the elapsed time between the SIM activation/change date and the current time. This endpoint is the dedicated signal for consumers (e.g., banks, fraud platforms) that need coarse time-band context for risk decisioning, ML features, or step-up logic. The consuming party applies its own thresholds outside the API contract. The response carries the band value only — it does NOT include the `swapped` Boolean (scope of `check`) or the change date (scope of `retrieve-date`), per the agreement reached in the 2026-04-29 ad-hoc call (issue #253) and TEF's stance on issue #260.
+
+      - Standardized AgeBand enum mapping: <br>
+        -1 = Subscriber valid, but no real-time profile found <br>
+         0 = Change has not happened <br>
+         1 = 0h ≤ d < 4h <br>
+         2 = 4h ≤ d < 12h <br>
+         3 = 12h ≤ d < 1d <br>
+         4 = 1d ≤ d < 2d <br>
+         5 = 2d ≤ d < 5d <br>
+         6 = 5d ≤ d < 7d <br>
+         7 = 7d ≤ d < 14d <br>
+         8 = 14d ≤ d < 30d <br>
+         9 = 30d ≤ d < 60d <br>
+        10 = 60d ≤ d < 90d <br>
+        11 = 90d ≤ d < 180d <br>
+        12 = 180d ≤ d < 1y <br>
+        13 = 1y ≤ d < 2y <br>
+        14 = 2y ≤ d < 3y <br>
+        15 = 3y+ <br>
 
     # Authorization and authentication
 
@@ -62,11 +86,11 @@ info:
 
     # Additional CAMARA error responses
 
-      The list of error codes in this API specification is not exhaustive. Therefore the API specification may not document some non-mandatory error statuses as indicated in `CAMARA API Design Guide`.
+    The list of error codes in this API specification is not exhaustive. Therefore the API specification may not document some non-mandatory error statuses as indicated in `CAMARA API Design Guide`.
 
-      Please refer to the `CAMARA_common.yaml` of the Commonalities Release associated to this API version for a complete list of error responses. The applicable Commonalities Release can be identified in the `API Readiness Checklist` document associated to this API version.
+    Please refer to the `CAMARA_common.yaml` of the Commonalities Release associated to this API version for a complete list of error responses. The applicable Commonalities Release can be identified in the `API Readiness Checklist` document associated to this API version.
 
-      As a specific rule, error `501 - NOT_IMPLEMENTED` can be only a possible error response if it is explicitly documented in the API.
+    As a specific rule, error `501 - NOT_IMPLEMENTED` can be only a possible error response if it is explicitly documented in the API.
 
     # Further info and support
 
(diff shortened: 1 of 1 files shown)


code/API_definitions/sim-swap.yaml
```diff
@@ -241,6 +343,76 @@ components:
...
+        - simSwapAgeBandEnum
+      description: |
+        Response schema for the /retrieve-age-band endpoint. Returns ONLY the standardized recency
+        band. This endpoint deliberately excludes the `swapped` Boolean (provided by `/check`)
+        and the SIM change timestamp (provided by `/retrieve-date`) so that each endpoint
+        contract stays single-purpose and unambiguous. Consuming parties apply their own risk
+        thresholds, ML features, or step-up logic against the returned band value outside the
+        API contract.
+      properties:
+        simSwapAgeBandEnum:
```
jpengar, line 414: It's a minor change, but I think we should keep the name of the property response field as 'SimSwapAgeBand'. The API specification already states that it is an enum.

code/API_definitions/sim-swap.yaml
```diff
@@ -241,6 +343,76 @@ components:
...
+    SimSwapAgeBandEnum:
+      type: integer
+      description: |
+        Standardized recency bucket representing the elapsed time between the SIM activation/change
+        date and the current time. The enum is fully standardized to ensure consistent interpretation
+        across operators, aggregators, and consuming applications.
+
+        | Value | Duration band                                         |
+        |-------|-------------------------------------------------------|
+        | -1    | Subscriber valid, but no real-time profile found      |
```
jpengar, line 355: As discussed during today's call, we should re-evaluate the '-1' use case. 

Firstly, in order to comply with the Commonalties guidelines, this should be modelled as an API error, e.g. 422, rather than mixing time bucket semantics with error scenarios. The integer values should be kept from 0 to 15.

Secondly, the error message should be generic enough to be applicable to any API provider; 'no real-time profile found' looks quite implementation-specific.

code/API_definitions/sim-swap.yaml
```diff
@@ -195,6 +225,78 @@ paths:
...
+      summary: Retrieve SIM swap age band
+      description: |
+        Returns a standardized `simSwapAgeBandEnum` recency bucket representing the elapsed time
+        between the SIM activation/change date and the current time.
+
+        This endpoint is the dedicated signal for consumers that need coarse time-band context
+        for fraud decisioning, risk scoring, ML features, or step-up logic. The response is
+        intentionally decoupled from `check` (Boolean threshold) and `retrieve-date` (timestamp)
+        to keep each endpoint contract clean and unambiguous. Accordingly, the response contains
+        the age band value ONLY — it does not include the `swapped` Boolean or the change date.
```
jpengar, line 249: The API should actually focus on providing network-related information, such as time buckets. How the API consumer then uses the data, e.g. as input for fraud scoring, would be up to them. The API just provides the information.

Avoid specifying the endpoint as a scoring signal. Instead, terms such as 'fraud decisioning', 'risk scoring', 'ML features' and 'step-up logic' are used. This is not neutral. This amounts to embedding business logic in the contract. We should make it more generic.

code/API_definitions/sim-swap.yaml
```diff
@@ -241,6 +343,76 @@ components:
...
+        | 6     | 5d ≤ d < 7d                                           |
+        | 7     | 7d ≤ d < 14d                                          |
+        | 8     | 14d ≤ d < 30d                                         |
+        | 9     | 30d ≤ d < 60d                                         |
+        | 10    | 60d ≤ d < 90d                                         |
+        | 11    | 90d ≤ d < 180d                                        |
+        | 12    | 180d ≤ d < 1y                                         |
+        | 13    | 1y ≤ d < 2y                                           |
+        | 14    | 2y ≤ d < 3y                                           |
+        | 15    | 3y+                                                   |
```
jpengar, line 371: The discussion about time buckets vs absolute timestamps is pending closure CC @HuubAppelboom

code/API_definitions/sim-swap.yaml
```diff
@@ -241,6 +343,76 @@ components:
...
+
+        Special states use negative and zero values so the band signal is compact and
+        self-contained: `0` means no SIM change has happened, and `-1` distinguishes a valid
+        subscriber with unavailable real-time profile data from both "no change" (`0`) and the
+        actual recency buckets (`1`–`15`).
+
+        NOTE: A newer CAMARA Commonalities release introduces a business-outcome modeling pattern
```
jpengar, line 378: Design decisions of this kind should be PR comments rather than being explicitly documented in the API spec.

jpengar (commented):
See Telefónica's comments below.

code/API_definitions/sim-swap.yaml
```diff
@@ -241,6 +343,76 @@ components:
...
+      type: integer
+      description: |
+        Standardized recency bucket representing the elapsed time between the SIM activation/change
+        date and the current time. The enum is fully standardized to ensure consistent interpretation
+        across operators, aggregators, and consuming applications.
+
+        | Value | Duration band                                         |
+        |-------|-------------------------------------------------------|
+        | -1    | Subscriber valid, but no real-time profile found      |
+        | 0     | Change has not happened                               |
```
jpengar, line 356: We may also re-evaluate the '0' use case. To be more consistent with the other two endpoints, we could define this field as nullable (`nullable: true`), returning `null` when a swap has not occurred. We could then just keep the enum integer values from '1' to '15'. WDYT?

CC @bigludo7

code/API_definitions/sim-swap.yaml
bigludo7: Agreed with @jpengar and meeting outcome is for @KeldaAnders to check in Commonalities the right message to apply. My guess is 422 but let see.

code/API_definitions/sim-swap.yaml
bigludo7: @jpengar excuse my lack of knowledge but we can have type `integer` and allow the attribute to be valued to `null`?

jpengar: @bigludo7 https://swagger.io/docs/specification/v3_0/data-models/enums/#nullable-enums

I think it could be something like this:

e.g.
```yaml
components:
  schemas:
    UserStatus:
      type: integer
      nullable: true
      enum:
        - 1
        - 2
        - 3
        - null
      description: |
        The status of the user:
        * `1` - Active
        * `2` - Inactive
        * `3` - Pending
        * `null` - No status assigned
```

PR #272, opened 2026-06-09
Update API definitions to fix CAMARA validation warnings and hints

bigludo7:
#### What type of PR is this?

Add one of the following kinds:

correction
cleanup
repository management

#### What this PR does / why we need it:

CAMARA validation is reporting many warnings and hints. This PR reduces that number.

#### Which issue(s) this PR fixes:

Fixes #271 

#### Changelog input

```
 release-note
 - Update API definitions to fix CAMARA validation warnings and hints
```

#### Additional documentation 

This section can be blank.

```
docs

```

PR #274, opened 2026-06-09
fix(release-plan): set target_api_status to alpha for the alpha pre-release

hdamker:
#### What type of PR is this?

correction

Sets `target_api_status` to `alpha` for both APIs in `release-plan.yaml`, matching `target_release_type: pre-release-alpha`. They were set to `public`, which declares GA maturity and applies the full public-status validation strictness — not appropriate for an alpha pre-release.

Fixes #

Release-plan only; no API definition changes.

#### Changelog input

```
 release-note
Correct release-plan target_api_status to alpha for the alpha pre-release
```

#### Additional documentation

This section can be blank.

```
docs

```

### DIFF
--- a/release-plan.yaml
+++ b/release-plan.yaml
@@ -34,14 +34,14 @@ dependencies:
 apis:
   - api_name: sim-swap-subscriptions
     target_api_version: 0.4.0
-    target_api_status: public
+    target_api_status: alpha
     main_contacts:
       - bigludo7
       - fernandopradocabrillo
       - maxl2287
   - api_name: sim-swap
     target_api_version: 2.2.0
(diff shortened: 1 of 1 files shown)


hdamker:
@camaraproject/sim-swap_codeowners: The PR should be re-validated after the fix https://github.com/camaraproject/tooling/pull/323 is available in v1-rc; will come back then and declare the PR "Ready for review".

hdamker:
The Spectral crash on null values is fixed. Reported validation errors are pre-existing (and are getting fixed by #272).

bigludo7 (approved):
LGTM

PR #275, opened 2026-07-16
Proposal for adding (optional) SIM type information

### LINKED ISSUE #241: Proposal for adding (optional) SIM type information
For API Consumers that want to process SIM Swap information as a protection against SIM Swap fraud, it may be beneficial to know what type of SIM Swap was performed, since there may exsist differences in the issueing proces between eSIM cards and physical SIM cards.

By adding the type of SIM card (physical SIM or eSIM) as an optional parameter to the SIM Swap API, API Consumers can differentiate according to the difference in risk associated with the different issueing process.

We propose to add both the type of the new SIM and the old SIM as optional parameters to be returned with the SIM Swap API, for example the parameters
- newSIMtype
- oldSIMtype

Both parameters need to be optional, and it should be decided by the API Provider whether they only want to share the newSIMtype, or both the newSIMtype and oldSIMtype.

What do you think of this enhancement ??

HuubAppelboom:
#### What type of PR is this?

* enhancement/feature

Customers that use the SIM Swap API in combination with Number Verify immediately after the user has migrated to a new device experience a  large number of false positives due to eSIMs.  By recognizing eSIMs and in particular the type of transfer with eSIMs, the API Consumer can correct for this.

Fixes #241 

@bigludo7 @jlurien @fernandopradocabrillo   Can you please review and approve this PR for fixing the false positives problem with eSIMs?

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -50,7 +50,7 @@ info:
       - when assuredTransfer=true: the CSP knows the SIM swap to an eSIM was performed through a CSP-verified transfer flow requiring possession/control of the previously active SIM or trusted device, for example by Apple eSIM Quick Transfer or Android eSIM Transfer functionality
       - when assuredTransfer=false: the CSP knows the SIM swap was not performed through such an assured transfer flow, for example after the SIM was lost or stolen.
       - if the CSP cannot determine this or does not support the attribute, the field should be omitted.
- 
+
     - POST retrieve-age-band : Returns a standardized `simSwapAgeBand` value indicating how recently a SIM swap occurred, expressed as a time band. This operation is an alternative way to expose SIM swap recency for API providers that do not expose the exact SIM swap date; it does not return the actual SIM swap date. This operation is OPTIONAL. A provider that does not implement it returns `501 NOT_IMPLEMENTED`; consumers can then fall back to `check` and/or `retrieve-date`. The returned value is a technical network signal indicating recency; it is not a customer-side risk score or scoring model. Consuming parties apply their own decisioning outside the API contract.
 
       - Definition of `d`:
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -194,12 +194,14 @@ paths:
               schema:
                 $ref: "#/components/schemas/SimSwapInfo"
               examples:
-                RETRIEVE_DATE:
-                  $ref: "#/components/examples/RETRIEVE_DATE"
+                RETRIEVE_DATE_WITHOUT_SYM_TYPE_AND_TRANSER_INFORMATION:
+                  $ref: "#/components/examples/RETRIEVE_DATE_WITHOUT_SYM_TYPE_AND_TRANSER_INFORMATION"
                 RETRIEVE_MONITORED_PERIOD:
                   $ref: "#/components/examples/RETRIEVE_MONITORED_PERIOD"
                 RETRIEVE_MONITORED_NULL:
                   $ref: "#/components/examples/RETRIEVE_MONITORED_NULL"
+                RETRIEVE_DATE_WITH_SYM_TYPE_AND_TRANSER_INFORMATION:  
+                  $ref: "#/components/examples/RETRIEVE_DATE_WITH_SYM_TYPE_AND_TRANSER_INFORMATION"
         "400":
           $ref: "../common/CAMARA_common.yaml#/components/responses/Generic400"
         "401":
--- a/code/API_definitions/sim-swap.yaml
+++ b/code/API_definitions/sim-swap.yaml
@@ -12,7 +12,7 @@ info:
 
     The SIM Swap API can also be used to protect non-automated actions. For example, when a call center expert contacts a user to clarify or confirm a sensitive operation.
 
-    This API is used by an application to get information about a mobile line latest SIM swap date. It can be easily integrated and used through this secured API and allows SPs (Service Provider) to get this information an easy & secured way. The API provides management of 3 endpoints answering 3 distinct questions: 
+    This API is used by an application to get information about a mobile line latest SIM swap date. It can be easily integrated and used through this secured API and allows SPs (Service Provider) to get this information an easy & secured way. The API provides management of 3 endpoints answering 3 distinct questions:
     An API provider is not expected to support all of them: the age-band operation is an alternative to exposing the exact SIM swap date, and support depends on the provider's available capabilities and commercial use case.
 
     * When did the last SIM swap occur?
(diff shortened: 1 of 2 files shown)


bigludo7:
Hello
I understood from #241 that we have an agreement for this one.

I'm just a bit anxious the calendar regarding freezing Syn26 release.
@HuubAppelboom could you solve the 2 blocking points? & @albertoramosmonagas I guess we're still on track to integrate this one for syn 26 but time running short right?

albertoramosmonagas:
Hi @HuubAppelboom, @bigludo7, some points from my side: 
1. The field should be named `simType` to comply with the CAMARA naming guidelines.
```yaml
simType:
  type: string
  maxLength: 4
  enum:
    - eSIM
    - pSIM
  description: The type of SIM that is currently active, either an eSIM or a physical SIM card.
```
2. The concept of `directTransfer` should be described more precisely. The current definition is quite generic and does not clearly explain what constitutes a direct transfer or how this process is performed from an implementation perspective.
3. Please replace the term **"Telecom Provider"** with **"Telco Operator"** throughout the proposal to maintain consistency with CAMARA terminology.

With these changes, I think we can move foward the PR

HuubAppelboom:
@bigludo7 @albertoramosmonagas 
The CLA authorization should be under way, hopefully today.

HuubAppelboom:
@albertoramosmonagas 
About the directTransfer: I agree with your comment that we should make it more precise, I am a bit puzzled how to do that.
Are we allowed to put links in the YAML towards for example urls where Apple or Android describe their versions of the directTransfer process ?  Like a link to https://support.apple.com/en-us/118669 ?
Are we even allowed to mention IOS and Android ?

Regarding the name directTransfer, we could also call this directDeviceTransfer

And in stead of a boolean, use an enum with directDeviceTransfer and carrierActivation
That may be more future proof in case we want to make it more granular in future.

What do you think ?

albertoramosmonagas:
Hi @HuubAppelboom, thanks for checking.

On links / vendor references: my understanding is that they are not forbidden, but the YAML should be self-contained for all normative behavior and should not depend on mutable third-party pages. So Apple / Android references could be mentioned as illustrative examples or kept in maintained explanatory documentation, but the API contract itself should remain vendor-neutral and not make those links normative.

On the field itself, I would avoid an enum such as `directDeviceTransfer` / `carrierActivation` for now, as that starts exposing transfer / activation mechanisms again and may become difficult to standardize across operators. I would prefer to keep it simple for this PR with an optional boolean field. I still think `directDeviceTransfer` is too narrow if we also want to cover same-device pSIM-to-eSIM migration. Something like `assuredTransfer` seems more accurate.

Suggested semantics:
* `assuredTransfer=true`: the SIM swap was performed through a CSP-verified transfer flow requiring possession/control of the previously active SIM or trusted device.
* `assuredTransfer=false`: the CSP knows the swap was not performed through such an assured transfer flow.
* if the CSP cannot determine this or does not support the attribute, the field should be omitted.

Both `simType` and `assuredTransfer` should remain optional and bound to the same SIM swap event the response refers to: the latest swap within the `/check` `maxAge` window, or the event behind `latestSimChange` for `/retrieve-date`.

So my suggestion would be: optional `simType` + optional `assuredTransfer`, vendor-neutral normative wording in the YAML, and any Apple/Android details only as non-normative examples or in maintained documentation.

HuubAppelboom:
@bigludo7 @albertoramosmonagas @fernandopradocabrillo   I have made all the requested changes, can you please review ?

albertoramosmonagas:
Hi @HuubAppelboom, thanks for the update.

The only blocking points I still see are:
1. `simType` should not be described as the SIM type that is “currently active”, as this may be ambiguous in multi-SIM scenarios. It should be explicitly bound to the relevant SIM swap event:
   * for `/check`, the latest swap within the requested `maxAge` window;
   * for `/retrieve-date`, the event behind `latestSimChange`.
2. The expected behaviour for `assuredTransfer` should be clarified:
   * `assuredTransfer=true`: the CSP knows the SIM swap was performed through a CSP-verified transfer flow requiring possession/control of the previously active SIM or trusted device.
   * `assuredTransfer=false`: the CSP knows the SIM swap was not performed through such an assured transfer flow.
   * if the CSP cannot determine this or does not support the attribute, the field should be omitted.

With these two clarifications, the proposal looks good from my side.

code/API_definitions/sim-swap.yaml
bigludo7, line 671: Should be renamed RETRIEVE_DATE_WITHOUT_SYM_TYPE_AND_TRANSER_INFORMATION

bigludo7, line 681: Add a fourth example RETRIEVE_DATE_WITH_SYM_TYPE_AND_TRANSER_INFORMATION

bigludo7 (commented):
Hello @HuubAppelboom 
Thanks for the contribution.
2 points for your consideration:
- Examples in the yaml need to be updated.
- As the API is stable we have to be precise in the Test definition. I guess we have to provide additional TCs in the sim-swap-retrieveSimSwapDate.feature file to test specifically these 2 new attributes: 1 test for eSIM, 1 for Physical, 1 for Transfert done by the operator, 1 for transfer not done by the operator. These tests are not mandatory and applies only if the carrier provide this information but for me they must be present.

HuubAppelboom:
@bigludo7 I added a fourth example showing a case of extra provided information. Can you check this one ?

@albertoramosmonagas I added some extra parts in de definition section. I also replaced "currently active" by the SIM after the swap event.  

Somehow the CAMARA Validation also fails. Is that normal ?

HuubAppelboom:
@bigludo7  I have als added 2TCs in the sim-swap-retrieveSimSwapDate.feature file, one for eSIM and assuredTransfer true, and one for pSIM and assuredTransfer false.  That probably is sufficient, or do we need to provide all possible combinations?
Can you plscheck these ?  I am not that familiar with these test case descriptions
Thanks in advance,

albertoramosmonagas:
Hi @HuubAppelboom, thanks for the update. A few points: 

1. On the CAMARA validation failure: this does not appear to be caused by the functional proposal itself, or by adding `simType` and `assuredTransfer`. It looks mainly like a specification consistency / YAML structure issue introduced by the latest changes.

   In particular, the `RETRIEVE_DATE` example was renamed, but an existing reference still points to the previous name, so the validator cannot resolve that component. The YAML also contains several trailing spaces and what looks like an indentation issue. Since OpenAPI components depend on the YAML hierarchy being exact, that indentation problem may also be causing the validator to interpret some sections incorrectly and report additional unresolved references, such as `x-correlator` and `ErrorInfo`.

2. The new retrieve-date test cases should probably not use `latestSimChange = null`. Since `simType` and `assuredTransfer` are bound to the SIM swap event behind `latestSimChange`, the test should cover a case where `latestSimChange` is actually returned. If the date is `null` because the event is outside the monitored / retention period, we should not expose event-level attributes for that non-returned event.

3. Since the new optional attributes are added to both `/retrieve-date` and `/check`, I think we should also add at least basic test coverage for `/check`, not only for `sim-swap-retrieveSimSwapDate.feature`.

On the number of combinations: from my side, I do not think we need to test all possible combinations. Two positive optional-information cases should be enough if they cover both `simType` values and both `assuredTransfer` values, for example `eSIM + assuredTransfer=true` and `pSIM + assuredTransfer=false`.

HuubAppelboom:
@bigludo7 HI Ludo, can you help with this PR ?  I am getting errors related to x-correlator in this PR in the RetrieveAgeBand area,  

The CAMARA file I could inspect shows /retrieve-age-band using local component references, while /retrieve-date and /check still use the shared common CAMARA reference.
 
Do you know what the correct refencing should be ? Should I change this in the YAML?  Or is this already being addressed in 
https://github.com/camaraproject/SimSwap/pull/278 ?

bigludo7:
Hi @HuubAppelboom  - for me it is fixed in #278 
I merge it and let see

bigludo7 (approved):
LGTM - Thanks @HuubAppelboom

HuubAppelboom:
@fernandopradocabrillo @maxl2287 Can you please review this PR as well?

hdamker:
@camaraproject/sim-swap_codeowners SimSwap is a repository with a stable API but no M3 release for Sync26 yet (overdue). My proposal is to change this PR here to draft and first create the M3 release candidate.

bigludo7:
What do you think @fernandopradocabrillo  and @HuubAppelboom ?
From Orange it was OK to add this for this release but I do not know from other companies....

albertoramosmonagas:
Hi @bigludo7, on behalf of Telefónica, we have no problem continuing with this PR, but we believe the milestones for the meta-release are too tight, and it would be better to leave it out so we can focus on the upcoming deadlines.

